PR #35, opened 2018-11-06
global: license change to MIT License

Fr4nk0is:


zzacharo:
@Fr4nk0is also docs folder should have the license I think!

zzacharo:
@Fr4nk0is LGTM. Request also from others to double check and before deploying to merge it. Is it possible to make it a script each time we merge on master to run it?

PR #51, opened 2019-04-09
searchapi: aggregator changes

vladbmihaescu:
* refactored code that displays nested aggregations

Screenshot from 2019-04-09 18-10-56

### DIFF
--- a/demo/src/index.js
+++ b/demo/src/index.js
@@ -8,7 +8,7 @@
 
 import React, { Component } from 'react';
 import { render } from 'react-dom';
-import { Segment, Container, Header, Grid } from 'semantic-ui-react';
+import { Segment, Container, Header, Grid, Accordion, Menu } from 'semantic-ui-react';
 import {
   ReactSearchKit,
   SearchBar,
@@ -20,6 +20,42 @@ import {
 import { Results } from './Results';
 
 class Demo extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { activeIndex: -1 }
+  }
+
+  handleClick = (e, titleProps) => {
+    const { index } = titleProps
+    const { activeIndex } = this.state
+    const newIndex = activeIndex === index ? -1 : index
+
+    this.setState({ activeIndex: newIndex })
+  };
+
+  renderAccordionAggregations = (
+    title,
+    resultsAggregations,
+    aggregations,
+    customProps
+  ) => {
+    const { activeIndex } = this.state;
+
+    return resultsAggregations !== undefined ? (
+      <Accordion as={Menu} vertical>
+        <Menu.Item>
+          <Accordion.Title
+            content={title}
+            index={customProps.index}
+            active={activeIndex===customProps.index}
+            onClick={this.handleClick}
+          />
+          <Accordion.Content active={activeIndex===customProps.index} content={aggregations} />
+        </Menu.Item>
+      </Accordion>
+    ) : null;
+  }
+
   render() {
     const sortByValues = [
       {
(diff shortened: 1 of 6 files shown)


src/api/SearchApi.js
```diff
@@ -74,39 +65,53 @@ class RequestSerializer {
...
-      key: bucket.key,
-      total: bucket.doc_count,
-      hasNestedField: false,
-    };
+const extractFromPath = (aggregation, pathToExtract) => {
+  return pathToExtract.reduce((obj, key) =>
+    (obj && obj[key] !== 'undefined') ? obj[key] : undefined, aggregation);
+}
+
+const createAggregation = (parentKeyName, buckets) => {
```
ntarocco: What do you think if you encapsulate both functions inside `ResponseSerializer` class?

src/api/SearchApi.js
```diff
@@ -74,39 +65,53 @@ class RequestSerializer {
...
+      aggregations[item.key] = createAggregation(newKeyName, nestedBuckets, nestedAggregationPath);
+      aggregations[item.key]["key"] = newKeyName;
+      aggregations[item.key]["name"] = item.key;
+      aggregations[item.key]["total"] = item.doc_count;
+      aggregations[item.key]["hasNestedField"] = nestedField;
     }
-    return aggregation;
-  }
+    else {
+      delete item.subtype;
```
ntarocco: Does this `item.subtype` always exists?

src/api/SearchApi.js
```diff
@@ -74,39 +65,53 @@ class RequestSerializer {
...
+      aggregations[item.key] = item;
```
ntarocco: I think you can delete this line no? Why not just:

```
const item = {
    'hasNestedField': false,
    'name': item.key,
    'total': item.doc_count,
    'key': parentKeyName.concat(".", item.key)
}
aggregations[item.key] = item;
```
But I might miss something here... the values maybe?

src/components/Aggregator/AggregatorValues.js
```diff
@@ -20,6 +20,47 @@ export default class AggregatorValues extends Component {
...
+  buildAggregations = (fieldName, values, userSelection) => {
+    const parentNestedField = values.hasNestedField;
+
+    return Object.keys(values).map((key, index) => {
+      if (typeof values[key] === 'object') {
+        const aggrValue = values[key];
+        const checked = this.isItemChecked(fieldName, aggrValue, userSelection);
+        const _onUserSelectionChange = (e, { value }) => {
+          const aggrNameValue = {};
+          const aggTypeName = parentNestedField ? parentNestedField: fieldName; // assuming the invenio bug where sub-aggregations with the same name have one type of url request
```
ntarocco: You cannot assume which backend you are using, the UI doesn't know it :)
I would re-phrase to something like:
`assuming that root fields names are unique`?

src/api/SearchApi.js
vladbmihaescu: I missed this, yes, now it deletes based on what the sub aggregation name is.

src/components/Aggregator/AggregatorValues.js
vladbmihaescu: I should've removed the line, it's not used anymore, now the keys are unique so it's not needed.

src/api/SearchApi.js
```diff
@@ -76,38 +76,54 @@ class RequestSerializer {
...
 /** Default backend response serializer */
 class ResponseSerializer {
-  _serializeAggregation(bucket) {
-    const aggregation = {
-      key: bucket.key,
-      total: bucket.doc_count,
-      hasNestedField: false,
-    };
+  _serializeAggregations(aggregationsResponse) {
+      const createAggregation = (parentKeyName, buckets) => {
```
zzacharo: I think split that to a `_ createAggregation` function would be more readable... WDYT?

src/api/SearchApi.js
```diff
@@ -76,38 +76,54 @@ class RequestSerializer {
...
+              "name": item.key,
+              "total": item.doc_count,
+              "key": "".concat(parentKeyName, ".", item.key)
+          };
+        }
+      });
+      return aggregations;
+    }
+
+    const extractFromPath = (aggregation, pathToExtract) => {
```
zzacharo: Maybe this one also?

src/api/SearchApi.js
vladbmihaescu: I agree, done!

demo/src/index.js
```diff
@@ -56,7 +92,7 @@ class Demo extends Component {
       <div>
         <ReactSearchKit
           searchConfig={{
-            url: 'https://zenodo.org/api/records/',
+            url: 'https://zenodo.org/api/records/',//,'https://localhost:5000/api/documents/'
```
ntarocco: @vladbmihaescu can you please remove this comment?

vladbmihaescu: Done!

PR #52, opened 2019-04-24
Searchapi modications

vladbmihaescu:


ntarocco:
@vladbmihaescu is this different from what you have already merged? It looks the same to me!

PR #54, opened 2019-05-05
wip: es integration

ntarocco:


ntarocco:
@kprzerwa this is a draft that also contains what you need, the URL Params enable/disable. If you need it urgently, you can continue this maybe, otherwise I will when I have some time.
I was planning to also add the initial query when you render, but not yet done...

Issue #59, opened 2019-06-12
Allow more customization on the search bar component


ppanero:
My goal is to implement the "suggest-as-you-type" on my UI. However the "onInputChange" listener is not "parametrized" on the search bar component.

To implement this, I'd have to overwrite the whole "renderElement" function. But I am happy with how it looks, so i'd just copy-paste and then loose possible updates. Is there a way/would it be useful to make the action functions configurable?

PR #60, opened 2019-06-20
components: add autocomplete search bar

ppanero:
Adds a new component: A search bar that, after more than 2 characters have been written to the search bar, will query for suggestions to help the user auto-complete his/her search.

The user has to:
- Configure the ``queryField`` against which the suggestions will be done (Typically a different field with an n-gram analyzer).
- Configure the ``responseField`` from which the suggestions will be taken from (In case it is different from the query).
- Optionally the ``Request`` and ``Response`` and serializers can be configured. 

The suggestions are saved in the global Redux state.

I tested in my app with both ``AutocompleteSearchBar`` and ``SearchBar`` and both work after the changes.

Might need some CSS changes:

- The search bar is ``border-radius`` but when the suggestions drop the corners do not match.
- The dropdown should end where the search button starts. Due to the nature of the SUIR ``Input`` component this is complicated.

closes https://github.com/inveniosoftware/react-searchkit/issues/59


ppanero:
@ntarocco Concerning https://github.com/inveniosoftware/react-searchkit/pull/54 please note that in this PR I created a new object ``InvenioSuggestionAPI`` that extends ``SearchApi``, just need to change one word but in case this gets merged before the WIP.

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
```diff
@@ -0,0 +1,161 @@
...
+        />
+        { this.renderSuggestions(querySuggestions) }
+      </div>
+    );
+  }
+
+  _handleAutocompleteChange = (queryString, debounce, debounce_time) => {
+    if(queryString.length >= 3) {
+      if (debounce){
+        const debouncedSuggestions = _.debounce(
```
ntarocco: May I suggest to define the function once to avoid to redefine the same each time?

There is no super nice solution, given that 1 param come from props. I see 2 options:
* a partial function
* define it in the constructor

Something like (a bit ugly?):
```
_suggest = value => (
    (debounceTime) => (
        _debounce(this...., debounceTime, { ... })
   )
)

_handleAutocompleteChange = (...) => {
    ...
    _suggest(value)(props.debounceTime);
}
```

Or constructor:
```
constructor(props) {
    ...
    this.suggest = _debounce(
        this.props.updateSuggetions,
        this.props.debounceTime,
        { leading: true, }
    )
}
```

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
```diff
@@ -0,0 +1,161 @@
...
+    
+    this.handleAutocompleteChange(queryString, debounce, debounce_time)
+  };
+
+  executeSearch = () => {
+    this.updateQueryString(this.state.currentValue);
+  };
+
+  render() {
+    const { placeholder, debounce, debounce_time, suggestions } = this.props;
```
ntarocco: debounce_time => debounceTime

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
```diff
@@ -0,0 +1,161 @@
...
+    );
+  }
+}
+
+AutocompleteSearchBar.propTypes = {
+  queryString: PropTypes.string,
+  updateQueryString: PropTypes.func.isRequired,
+  updateSuggetions: PropTypes.func.isRequired,
+  clearSuggestions: PropTypes.func.isRequired,
+  renderElement: PropTypes.func,
```
ntarocco: I think here it is missing debounce and debounceTime params (optional or required), and a default value below.

src/state/actions/query.js
```diff
@@ -157,3 +160,45 @@ export const executeQuery = (updateUrlParams = true) => {
     }
   };
 };
+
+export const updateSuggestions = suggestionString => {
+  return async dispatch => {
+    dispatch({
+      type: SET_SUGGESTION_STRING,
```
ntarocco: I have the feeling that you don't need to store in the `query` state the suggestion string. We use the state to persist all user selection, in the sense that if you for example modify the sort by or pagination, we want to keep the same query and sort order previously selected by the user.
I think here you perform always a query when asking for suggestion, you don't need any other previous or other value, no?

But again I might be wrong! :)

src/api/SuggestionApi.js
```diff
@@ -0,0 +1,83 @@
...
+
+  _extractSuggestion(hit) {
+    var currentObject = hit
+
+    this.responseFieldPath.forEach((field) => {
+      currentObject = currentObject[field]
+    });
+
+    return currentObject;
+  }
```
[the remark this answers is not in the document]
ppanero: Not really, because the path can be nested. E.g.

``` json
[
    { 
        "a": {
            "b": {
                "title": "value to extract"
            }
        }
   },
  { 
      "a": {
          "b": {
              "title": "value to extract"
          }
      }
  }
]
```

Therefore I have to loop in the object until I react the inner most (``title`` in the example). It can also not be accessed normally because we cannot assume how many levels of nesting might there be.

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
ppanero: Ok for the missing type. As for the default, i was defining it the same way than the placeholder:

``` JavaScript
    // Initialize values
    placeholder = placeholder || 'Type something';
    debounce = debounce || false;
```

In ``_renderElement``, that same way that the ``SearchBar`` component does. Should this be changed?

src/api/SuggestionApi.js
```diff
@@ -0,0 +1,83 @@
...
+
+/** Default suggestions response serializer */
+class InvenioSuggestionResponseSerializer {
+
+  constructor(responseField) {
+    this.responseFieldPath = responseField.split(".");
+  }
+
+  _extractSuggestion(hit) {
+    var currentObject = hit
```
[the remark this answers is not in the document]
ppanero: Needs reassignment, so not to be RO. Changed by ``let``, cant be ``const``.

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
ppanero: Moved all to the constructor. Since ``debounce`` and ``debounceTime`` are in "props" it can all be just taken in the constructor and the conditional assignment done there. Therefore, it also simplifies the handler function.

src/api/SuggestionApi.js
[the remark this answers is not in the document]
zzacharo: @ppanero I think this works for you [lodash/get](https://lodash.com/docs/4.17.11#get) no?

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
zzacharo: @ppanero would be nice to deload the code and move defaults to defaultProps :) Start with your new component ;)

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
```diff
@@ -0,0 +1,150 @@
...
+      currentValue: queryString,
+    });
+
+    this.handleAutocompleteChange();
+  };
+
+  executeSearch = () => {
+    this.updateQueryString(this.state.currentValue);
+  };
+
```
zzacharo: I just had an order issue in the method declarations, probably my ocd :P, could you move the `_renderElement` here?

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
```diff
@@ -0,0 +1,150 @@
...
+      this.updateSuggetions(this.state.currentValue);
+    }
+  };
+
+  onInputChange = async queryString => {
+    await this.setState({
+      currentValue: queryString,
+    });
+
+    this.handleAutocompleteChange();
```
zzacharo: How have you imagined that to be used from user when they override it? In general I think we have this problem that we expose props that we imply that if you override you need to execute some piece of code that was already there, e.g update component state/redux state, and this is not user friendly. I think, as I understood from what you are doing here, a prop `minCharsToAutocomplete` should be enough.

src/api/SuggestionApi.js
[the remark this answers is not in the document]
ppanero: Yeay! Thanks!

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
ppanero: done

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
ppanero: done

src/components/AutocompleteSearchBar/AutocompleteSearchBar.js
ppanero: Done. Passed the querySuggestion as parameter to ease the overload.

src/state/actions/query.js
ppanero: As discussed IRL it is usefull to save to redux e.g. in case of filtering by aggregations.

ppanero:
@ntarocco @zzacharo Thanks for the review! All changes have been implemented :D

PR #64, opened 2019-07-14
global: js dep upgrade + optional url persistency

ntarocco:
* upgraded all JS dependencies to latest
* added eslint
* user can now disable URL Query String persistence


ntarocco:
I need to add more tests, but I would like to have your feedback in the meanwhile.

PR #69, opened 2019-08-09
global: replace nwb with create-react-app

ntarocco:
- nwb is not anymore mantained and it is obsolete with many security
issues
- create-react-app is now used for development
- babel cli is used for building the library, but is not optimal. It
will require the integration of a build system with ad-hoc configuration


ntarocco:
@zzacharo the AutocompleteSearchBar has some `css`, it might break for that.
I think we have to find another way to build the lib, with a small webpack or any other build system.
Maybe we can merge this, and then see how to do it.

PR #77, opened 2019-10-06
api: add ElasticSearch API

ntarocco:
* adds ElasticSearch API and related demo example
* allows overriding of ResultsList and ResultsGrid components in
  ResultsMultiLayout component
* upgrades react-scripts


src/lib/api/contrib/elasticsearch/ESRequestSerializer.js
```diff
@@ -0,0 +1,79 @@
...
+    if (size > 0) {
+      bodyParams['size'] = size;
+    }
+
+    if (page > 0) {
+      const s = size > 0 ? size : 0;
+      const from = (page - 1) * s;
+      bodyParams['from'] = from;
+    }
+    // this._addAggregations(getParams, aggregations);
```
zzacharo: leftover?

ntarocco: it will come in the next PR. I will remove it then if you prefer.

zzacharo: No it's fine by me!

src/demos/elasticsearch/App.js
```diff
@@ -0,0 +1,69 @@
...
+
+const searchApi = new ESSearchApi({
+  url: 'http://localhost:5000/random/_search',
+  timeout: 5000,
+});
+
+export class App extends Component {
+  constructor(props) {
+    super(props);
+  }
```
topless: This is not needed.

src/demos/elasticsearch/Results.js
```diff
@@ -0,0 +1,113 @@
...
+    default:
+      style = { margin: `0 ${size}` };
+  }
+  return <span style={style}>{text}</span>;
+};
+
+export class Results extends Component {
+  constructor(props) {
+    super(props);
+  }
```
topless: Same here

src/demos/zenodo/Results.js
```diff
@@ -0,0 +1,156 @@
...
+        </Grid>
+        <Grid relaxed verticalAlign="middle" textAlign="center">
+          <Pagination />
+        </Grid>
+      </div>
+    ) : null;
+  }
+}
+
+Results.propTypes = {};
```
topless: Shouldn't `sortValues` and `resultsPerPageValues` be a part of that?

src/demos/elasticsearch/App.js
ntarocco: There will be more in the next PR for this, that's why I kept. If it will be not needed, I will remove, thanks for checking.

src/demos/elasticsearch/Results.js
ntarocco: Same as above

src/demos/zenodo/Results.js
ntarocco: Yes they should, but this is a demo example, not really the lib, I am not sure it is so important :)

PR #78, opened 2019-10-14
pagination: allow passing props

kprzerwa:
* enables passing semantic-ui props

### DIFF
--- a/src/lib/components/Pagination/Pagination.js
+++ b/src/lib/components/Pagination/Pagination.js
@@ -38,7 +38,8 @@ export default class Pagination extends Component {
     currentSize,
     totalResults,
     onPageChange,
-    options
+    options,
+    ...extraParams
   ) => {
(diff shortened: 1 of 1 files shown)


src/lib/components/Pagination/Pagination.js
```diff
@@ -38,7 +38,8 @@ export default class Pagination extends Component {
     currentSize,
     totalResults,
     onPageChange,
-    options
+    options,
+    ...restParams
```
ntarocco: nit: `restParams` -> `otherParams`/`extraParams`? It reminds me of REST APIs :relaxed:

for the `nit`, in case you did not read the [article](https://stackoverflow.blog/2019/09/30/how-to-make-good-code-reviews-better/)...

kprzerwa: I've read it:P Thanks !

PR #89, opened 2020-01-17
searchApi: fix "url" axios config override

### LINKED ISSUE #88: InvenioSearchApi: accidental use of both "baseURL" and "url" config
axios supports [both `url` and `baseURL` parameters](https://github.com/axios/axios#request-config), setting an absolute `url` though (i.e. `url: "https://localhost:5000/api/communities"`, cancels `baseURL`.

Setting [`baseURL: config.url`](https://github.com/inveniosoftware/react-searchkit/blob/001592c8edea81db429ae1be5bbb1d42c3998671/src/lib/api/contrib/invenio/InvenioSearchApi.js#L48) and then expanding the config via `...config`, is the equivalent of:

```javascript
const axiosConfig = {
  baseURL: config.url,
  url: config.url,
};
```

When a relative `url` is used (e.g. `url: "/api/communities"`), this causes the final requests to be made to `baseURL + url = "/api/communities/api/communities"`.

slint:
* Closes #88.

### DIFF
--- a/src/lib/api/contrib/elasticsearch/ESSearchApi.js
+++ b/src/lib/api/contrib/elasticsearch/ESSearchApi.js
@@ -14,15 +14,22 @@ import { ESResponseSerializer } from './ESResponseSerializer';
 
 export class ESSearchApi {
   constructor(config) {
-    this.validateConfig(config);
+    this.axiosConfig = _get(config, 'axios', {});
+    this.validateAxiosConfig();
     this.initSerializers(config);
-    this.initAxios(config);
+    this.initInterceptors(config);
+    this.initAxios();
   }
 
-  validateConfig(config) {
-    if (!_hasIn(config, 'url')) {
+  validateAxiosConfig() {
+    if (!_hasIn(this.axiosConfig, 'url')) {
       throw new Error('ESSearchApi config: `node` field is required.');
     }
+  }
+
+  initInterceptors(config) {
+    this.requestInterceptor = _get(config, 'interceptors.request', undefined);
+    this.responseInterceptor = _get(config, 'interceptors.response', undefined);
   }
 
   initSerializers(config) {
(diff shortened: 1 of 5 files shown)


src/lib/api/contrib/elasticsearch/ESSearchApi.js
```diff
@@ -41,12 +41,9 @@ export class ESSearchApi {
     this.responseSerializer = new responseSerializerCls();
   }
 
-  initAxios(config) {
+  initAxios({ url, ...config }) {
```
zzacharo: @slint How is this different than the existing implementation?
Before the code was expecting an object e.g `{url: '/me'}` and this would end up in the non desired behaviour of calling `/me/me` when it was triggering a request. Now with the change passing the passing the same object I think it would lead to the same outcome no? Unless you pass explicitly the `baseURL` parameter which in that case it was already possible before...Do I miss something?

slint: Since now `url` is a destructured value from the passed object function argument, it is will not be inside the `...config` object. This means that it won't be passed at all to the `axios.create(...)` call

zzacharo: But then you will not be able to pass the `url` parameter at all right?

slint: Yup. We have to discuss whether we'll directly pass-through the config to `axios.create(...)`, or if we'll introduce an `axiosConfig` parameter to the `ESSearchApi`/`InvneioSearchApi` class constructor.

zzacharo: @ntarocco what do you think? I would be in favor I think of name spacing the axios config like `axiosConfig` and pass it through to axios.

ntarocco: Are we discussing this? https://github.com/inveniosoftware/react-searchkit/issues/88
If yes, I should have already replied there :)

zzacharo: @ntarocco sorry I missed that. So just to confirm what you are proposing is just get through all the config passed to the axios instance right? In that case I wouldn't destructure the `url` parameter from the config passed and just pass through the config as it is. Then, it would be clear for the users (we should update the docs for that too) that this config is the same with what axios is accepting. @slint WDYT?

slint: I would go with an explicit `axiosConfig` parameter for the `InvenioSearchApi` constructor:

https://github.com/inveniosoftware/react-searchkit/blob/375f51db4b3859ddf809f97b513cc0f7e428c544/src/lib/api/contrib/invenio/InvenioSearchApi.js#L15-L16

So something like:

```javascript
export class InvenioSearchApi {
  constructor(serializers, axiosConfig, axiosInterceptors) {
    // NOTE: Would still validate the axiosConfig, but inside "initAxios"
    // this.validateConfig(config);
    this.initSerializers(serializers);
    this.initAxios(axiosConfig, axiosInterceptors);
  }
  ...
}
```

slint:
LGTM :shipit: (I can't review my own PR apparently)

PR #100, opened 2020-03-30
global: expose updateQueryState

### LINKED ISSUE #99: Allow queryState change in `withState` HOC component
When an app implements its own custom component using `withState` HOC component, currently it can only read the state, but not modify it. Add the action `updateQueryState` to it.

1. add the action `updateQueryState` in [withState](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/HOC/withState.js#L15) component, as done [here](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/Bootstrap/index.js#L19)
2. test the changes for filters: given an app with 2 selected filters, e.g. `file_type=pdf` and `language=en`, when the new custom components updates `language=it`,  `file_type=pdf` should be kept as is.
3. Add tests and update docs and demos if necesssary.

### LINKED ISSUE #101: bug: ResultsLoader children multiple instances
All the components that are wrapped inside `<ResultsLoader></ResultsLoader>` component are constructed twice check [ResultsLoader render](https://github.com/inveniosoftware/react-searchkit/blob/1f79c1b66ec5fecd52b028678329e1c36ba039a3/src/lib/components/ResultsLoader/ResultsLoader.js#L22).

This is due to the fact that our state starts with `loading = false`, so what happens is 

1) initial state `loading = false`: children get instantiated
2) loading state`loading = true`: spinner is shown all good
3) completed state `loading = false`: children get instantiated (again)

Although changing the [initialState](https://github.com/inveniosoftware/react-searchkit/blob/1f79c1b66ec5fecd52b028678329e1c36ba039a3/src/lib/store.js#L32) to `true` does the trick, a singleton on children or another approach might serve us better, @zzacharo @ntarocco WDYT?

topless:
- updateQueryState exposed withState HOC
- test for exposed props form withState
- documentation
- fix for config expecting axios property
- fix circular dependency inside `store.js` -> `reducers/index.js` -> `state/reducers/query.js` -> `store.js`
- changed initialResultsState.loading to true to avoid double children render in ResultsLoader closes #101
- closes #99

### DIFF
--- a/docs/docs/components/with_state.md
+++ b/docs/docs/components/with_state.md
@@ -3,11 +3,17 @@ id: with-state
 title: withState
 ---
 
-`withState` is a HOC component that passes the redux state to an external component.
+`withState` is a high order component, which is used to expose the redux state and
+actions to external components.
 
-The component receives the up-to-date state every time something is changed.
+The component receives the up-to-date state every time something is changed and
+through its props it gains access to
 
-> Do **not** mutate the state inside your wrapped component!
+- `currentResultsState`
+- `currentQueryState`
+- `updateQueryState`
+
+> Do **not** mutate directly the state inside your wrapped component. Instead, use the function `updateQueryState` to pass your new query.
 
 ## Usage
 
--- a/src/demos/cern-videos/App.js
+++ b/src/demos/cern-videos/App.js
@@ -48,8 +48,10 @@ const resultsPerPageValues = [
 ];
 
 const searchApi = new InvenioSearchApi({
-  url: 'https://videos.cern.ch/api/records/',
-  timeout: 5000,
+  axios: {
+    url: 'https://videos.cern.ch/api/records/',
+    timeout: 5000,
+  },
 });
 
 export class App extends Component {
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ React-SearchKit is a React library that allows you to build in an easy way your 
 
 Main features:
 
-* ready-to-use collection of UI components
-* configurable REST API endpoint and serialization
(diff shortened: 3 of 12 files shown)


docs/docs/components/with_state.md
```diff
@@ -3,11 +3,17 @@ id: with-state
...
-The component receives the up-to-date state every time something is changed.
+The component receives the up-to-date state every time something is changed and
+through its props it gains access to
 
-> Do **not** mutate the state inside your wrapped component!
+- `currentResultsState`
+- `currentQueryState`
+- `updateQueryState`
+
+> Do **not** mutate directly the state inside your wrapped component, used instead pass the function `updateQueryState` to pass your new query.
```
ntarocco:
```suggestion
> Do **not** mutate directly the state inside your wrapped component. Instead, use the function `updateQueryState` to pass your new query.
```

ntarocco (approved):
👍 LGTM!
Were you able to try if the state is mutate correctly in case of filters (previous filters are kept)?

src/demos/cern-videos/App.js
```diff
@@ -48,8 +48,10 @@ const resultsPerPageValues = [
 ];
 
 const searchApi = new InvenioSearchApi({
-  url: 'https://videos.cern.ch/api/records/',
-  timeout: 5000,
+  axios: {
```
zzacharo: @topless Thanks for fixing that I totally missed it :(

topless: @zzacharo no worries! I have fixed everything on ILS side too!

zzacharo (approved):

PR #102, opened 2020-04-01
global: updateQueryState support for multiple filters

topless:
`updateQueryState` accepts only one filter a time. The intention of this pr is to be able to submit multiple filters at once. This is useful for example when we want to replace a filter.

### DIFF
--- a/src/lib/state/selectors/query.js
+++ b/src/lib/state/selectors/query.js
@@ -7,8 +7,12 @@
  */
 
 import Qs from 'qs';
+import _forEach from 'lodash/forEach';
+import _head from 'lodash/head';
+import _isArray from 'lodash/isArray';
 import _isEmpty from 'lodash/isEmpty';
 import _pick from 'lodash/pick';
+
 /**
  * Return true if the first string starts and contains the second.
  * @param {string} first a string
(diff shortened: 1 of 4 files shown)


src/lib/state/selectors/query.js
```diff
@@ -101,6 +104,25 @@ export const updateQueryFilters = (queryFilter, stateFilters) => {
...
+  /** If we have one filter as query = ['file_type', 'pdf'] */
+  if (!_isArray(_head(queryFilter))) {
+    return updateFilter(queryFilter, stateFilters);
+  }
+
+  /** If we have an array of filters as query we apply the filters one by one.
+   * e.g. query = [['file_type', 'pdf'], ['file_type', 'txt']]
+   */
+  let tempStateFilters = stateFilters;
+  _forEach(
```
ntarocco, line 121: can you do simply:
```
return queryFilters.filter(queryFilter => updateFilter(queryFilter. stateFilters))
```
?

topless: This was my initial approach but as we call updateFilter our stateFilters change, so our next updateFilter should be applied on the altered state and not the initial one. I can explain in a call if needed.

PR #106, opened 2020-04-13
global: add prefix and suffix props

### LINKED ISSUE #104: Add "prefix" and "suffix" props to components
For some components it would be nice to pass a "prefix" and a "suffix" component so that the component itself can be rendered in the middle.
For example, for the component `ResultsPerPage` it would be nice to have:
```
<ResultsPerPage
  prefix={'Show '}
  suffix={' results per page'}
```
so that it then renders:
`Show 30 results per page`

On top of that, if the component does not render because there are 0 results, the prefix and suffix are also hidden. Currently it is not possible.

KonstantinaStoikou:
Closes #104

Example:
Selection_001

### DIFF
--- a/docs/docs/components/count.md
+++ b/docs/docs/components/count.md
@@ -21,6 +21,14 @@ The component is **not** displayed while executing the search query or if there 
 
   An optional function to override the default rendered component.
 
+- **prefix** `String` _optional_
+
+  An optional string to be displayed before the number of results.
+
+- **suffix** `String` _optional_
+
+  An optional string to be displayed after the number of results.
+
 ## Usage when overriding template
 
 ```jsx
--- a/docs/docs/components/count.md
+++ b/docs/docs/components/count.md
@@ -24,7 +24,7 @@ The component is **not** displayed while executing the search query or if there 
 - **label** `function` _optional_
 
   An optional function to wrap the component with a prefix and suffix string. <br />
-  E.g. `label={(val) => <> Found {val} results</>} />`
+  E.g. `label={(cmp) => <> Found {cmp} results</>} />`
 
 ## Usage when overriding template
 
--- a/docs/docs/components/count.md
+++ b/docs/docs/components/count.md
@@ -21,13 +21,10 @@ The component is **not** displayed while executing the search query or if there 
 
   An optional function to override the default rendered component.
 
-- **prefix** `String` _optional_
+- **label** `function` _optional_
 
-  An optional string to be displayed before the number of results.
-
-- **suffix** `String` _optional_
-
-  An optional string to be displayed after the number of results.
+  An optional function to wrap the component with a prefix and suffix string. <br />
+  E.g. `label={(val) => <> Found {val} results</>} />`
 
 ## Usage when overriding template
 
--- a/src/lib/components/Count/Count.js
+++ b/src/lib/components/Count/Count.js
@@ -23,12 +23,9 @@ export default class Count extends Component {
 
   render() {
     const { loading, totalResults, label } = this.props;
-    const size = '0.5em';
     return (
       <ShouldRender condition={!loading && totalResults > 0}>
-        <span style={({ marginLeft: size }, { marginRight: size })}>
-          {label(this.renderElement(totalResults))}
-        </span>
+        {label(this.renderElement(totalResults))}
       </ShouldRender>
     );
   }
@@ -43,5 +40,5 @@ Count.propTypes = {
 
 Count.defaultProps = {
   renderElement: null,
-  label: (val) => val,
+  label: (cmp) => cmp,
 };
--- a/src/lib/components/Count/Count.js
+++ b/src/lib/components/Count/Count.js
@@ -21,29 +21,14 @@ export default class Count extends Component {
     return <Label color={'blue'}>{totalResults}</Label>;
   }
 
-  renderSpanWithMargin = (text, margin) => {
-    const size = '0.5em';
-    let style;
-    switch (margin) {
-      case 'left':
-        style = { marginLeft: size };
-        break;
-      case 'right':
-        style = { marginRight: size };
-        break;
-      default:
-        style = { margin: `0 ${size}` };
-    }
-    return <span style={style}>{text}</span>;
-  };
-
   render() {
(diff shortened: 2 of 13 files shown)


src/lib/components/Count/Count.js
```diff
@@ -21,11 +21,29 @@ export default class Count extends Component {
...
+    }
+    return <span style={style}>{text}</span>;
+  };
+
   render() {
-    const { loading, totalResults } = this.props;
+    const { loading, totalResults, prefix, suffix } = this.props;
     return (
       <ShouldRender condition={!loading && totalResults > 0}>
+        {this.renderSpanWithMargin(prefix, 'right')}
```
ntarocco, line 44: Prefix or Suffix might be not defined, so the extra margin should not appear. I would change both to:
`{prefix && this.renderSpanWithMargin(prefix, 'right')}`

WDYT?

src/lib/components/ResultsPerPage/ResultsPerPage.js
```diff
@@ -49,13 +49,31 @@ export default class ResultsPerPage extends Component {
...
+  };
+
   render() {
-    const { loading, currentSize, totalResults } = this.props;
+    const { loading, currentSize, totalResults, prefix, suffix } = this.props;
     return (
       <ShouldRender
         condition={!loading && totalResults > 0 && currentSize !== -1}
       >
+        {this.renderSpanWithMargin(prefix, 'right')}
```
ntarocco, line 74: Same here

src/lib/components/ResultsPerPage/ResultsPerPage.js
```diff
@@ -49,13 +49,31 @@ export default class ResultsPerPage extends Component {
...
+        break;
+      default:
+        style = { margin: `0 ${size}` };
+    }
+    return <span style={style}>{text}</span>;
+  };
+
   render() {
-    const { loading, currentSize, totalResults } = this.props;
+    const { loading, currentSize, totalResults, prefix, suffix } = this.props;
```
zzacharo, line 69: May I suggest a change that IMHO it fits a bit better the purpose of the prefix/suffix props without making it too specific? As I understood, please correct me if I am wrong, the end goal is to apply a custom label that wraps the ResultsPerPage component. I would go then for a label prop that can be a function like this: 
```javascript
ResultsPerPage label={(cmp) => `Show {cmp} results per page`} />
```

then in the render function of the `ResultsPerPage` component you could write something similar to this:
```javascript

render() {
const { loading, currentSize, totalResults, label } = this.props;
...
{label && label(this.renderElement(currentSize, this.options, this.onChange))}
...
}
```

src/lib/components/ResultsPerPage/ResultsPerPage.js
```diff
@@ -49,31 +49,16 @@ export default class ResultsPerPage extends Component {
...
+    const { loading, currentSize, totalResults, label } = this.props;
+    const size = '0.5em';
     return (
       <ShouldRender
         condition={!loading && totalResults > 0 && currentSize !== -1}
       >
-        {prefix && this.renderSpanWithMargin(prefix, 'right')}
-        {this.renderElement(currentSize, this.options, this.onChange)}
-        {suffix && this.renderSpanWithMargin(suffix)}
+        <span style={({ marginLeft: size }, { marginRight: size })}>
```
zzacharo, line 59: I think this styling should go to the application that uses this component so as to keep the components here with minimal styling :)

src/lib/components/Sort/Sort.js
```diff
@@ -134,14 +118,16 @@ export default class Sort extends Component {
...
       >
-        {prefix && this.renderSpanWithMargin(prefix, 'right')}
-        {this.renderElement(
-          currentSortBy,
-          currentSortOrder,
-          this.options,
-          this.onChange
-        )}
-        {suffix && this.renderSpanWithMargin(suffix)}
+        <span style={({ marginLeft: size }, { marginRight: size })}>
```
zzacharo, line 121: Same here for the styling!

src/lib/components/SortBy/SortBy.js
```diff
@@ -53,31 +53,18 @@ export default class SortBy extends Component {
...
+    const { currentSortBy, loading, totalResults, label } = this.props;
+    const size = '0.5em';
     return (
       <ShouldRender
         condition={currentSortBy !== null && !loading && totalResults > 0}
       >
-        {prefix && this.renderSpanWithMargin(prefix, 'right')}
-        {this.renderElement(currentSortBy, this.options, this.onChange)}
-        {suffix && this.renderSpanWithMargin(suffix)}
+        <span style={({ marginLeft: size }, { marginRight: size })}>
```
zzacharo, line 63: Same here for the styling!

src/lib/components/SortOrder/SortOrder.js
```diff
@@ -49,37 +49,18 @@ export default class SortOrder extends Component {
...
+    const { currentSortOrder, loading, totalResults, label } = this.props;
+    const size = '0.5em';
     return (
       <ShouldRender
         condition={currentSortOrder !== null && !loading && totalResults > 0}
       >
-        {prefix && this.renderSpanWithMargin(prefix, 'right')}
-        {this.renderElement(currentSortOrder, this.options, this.onChange)}
-        {suffix && this.renderSpanWithMargin(suffix)}
+        <span style={({ marginLeft: size }, { marginRight: size })}>
```
zzacharo, line 59: Same here, sorry :)

docs/docs/components/count.md
```diff
@@ -21,13 +21,10 @@ The component is **not** displayed while executing the search query or if there
...
-- **prefix** `String` _optional_
+- **label** `function` _optional_
 
-  An optional string to be displayed before the number of results.
-
-- **suffix** `String` _optional_
-
-  An optional string to be displayed after the number of results.
+  An optional function to wrap the component with a prefix and suffix string. <br />
+  E.g. `label={(val) => <> Found {val} results</>} />`
```
zzacharo, line 27: I would probably change the `val` to `cmp` just to indicate that the passed argument is the component not just a value. This would make clear that the return value of the `label` function needs to be a component and cannot be a string :) Please do the corresponding changes to the other places :)

PR #116, opened 2020-05-11
global: mark components as Overridable

### LINKED ISSUE #114: renderElement function should be a component prop
The `renderElement` function should be converted to a component prop so as `ShouldRender` component render it correctly when the condition is true. Otherwise, the function is being evaluated i.e the component that it is returned without the respecting the condition's value.

### LINKED ISSUE #115: Mark components overridable via `Overridable` library
In each UI component:
- [x] change the overriding of default template rendering to use `Overridable`
- [ ] enable app and component name spacing
  - [ ] would it be possible to raise/throw if the same component is added 2 times and the `name/key/id` prop to namespace it is not passed or already defined?
- [x] make sure the `<ShouldRender` children are React components and not functions (see https://github.com/inveniosoftware/react-searchkit/issues/114)
- [x] should we remove the `renderElement` functions?
- [ ] make sure all demos are working
- [ ] update the docs

mvidalgarcia:
closes #115
closes #114


src/lib/components/BucketAggregation/BucketAggregation.js
```diff
@@ -85,13 +85,34 @@ BucketAggregation.propTypes = {
...
+  overridableUID: '',
+};
+
+const Element = (props) => {
+  const { title, containerCmp, overridableUID } = props;
+  return (
+    containerCmp && (
+      <Overridable
+        id={buildUID('BucketAggregation.element', overridableUID)}
+        {...props}
```
ntarocco, line 105: will it pass more `props` here? Shall we have a strict interface instead so that when you override you get only what you can use?
e.g.
```
const myNewComp = ({ title, containerCmp }) = ()
```
or maybe we can just documented it well and leave it like this...

src/lib/components/EmptyResults/EmptyResults.js
```diff
@@ -51,9 +48,34 @@ EmptyResults.propTypes = {
...
+    <Overridable
+      id={buildUID('EmptyResults.element', overridableUID)}
+      {...props}
+    >
+      <Segment placeholder textAlign="center">
+        <Header icon>
+          <Icon name="search" />
+          No results found!
+        </Header>
+        <em>Current search "{queryString}"</em>
```
ntarocco, line 71: There is a small bug in RSK, you can have 0 results but the queryString is empty. Could you please add:
```suggestion
        {queryString && <em>Current search "{queryString}"</em>}
```

src/lib/components/ReactSearchKit/ReactSearchKit.js
```diff
@@ -41,7 +43,9 @@ export class ReactSearchKit extends Component {
           appName={this.appName}
           eventListenerEnabled={this.eventListenerEnabled}
         >
-          {this.props.children}
+          <Overridable id={buildUID('ReactSearchkit.children', overridableUID)}>
```
ntarocco, line 46:
```suggestion
          <Overridable id={buildUID('ReactSearchKit.children', overridableUID)}>
```

src/lib/components/ActiveFilters/ActiveFilters.js
```diff
@@ -35,36 +34,57 @@ export default class ActiveFilters extends Component {
...
+      )
+    );
   }
 }
 
 ActiveFilters.propTypes = {
   filters: PropTypes.array.isRequired,
   updateQueryFilters: PropTypes.func.isRequired,
   renderElement: PropTypes.func,
+  overridableUID: PropTypes.string,
```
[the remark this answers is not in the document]
mvidalgarcia: I don't have a strong opinion, said that:
- `uniqueId`: I think is way too generic, I would include the "overridable" concept somehow as it's directly related to it.
- `overridableUniqueId`: I think it's too long.
- Both `overridableId` and `overidableUID` are good for me.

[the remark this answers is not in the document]
ntarocco: Let's go for `overridableId` and close the topic! :)

src/lib/components/BucketAggregation/BucketAggregation.js
mvidalgarcia: It's actually only passing `title` and `containerCmp`. We can filter out props when unpacking. I forgot to do it in this component but I usually filter out `overridableUID`.
```js
const Element = ({ overridableUID, ...props }) => {
```

src/lib/components/ReactSearchKit/ReactSearchKit.js
mvidalgarcia: Good catch 🕵️‍♂️

PR #128, opened 2020-08-03
components: extend Aggregation and EmptyResults interface

kprzerwa:


src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -127,7 +138,7 @@ const ValueElement = (props) => {
     >
       <List.Item key={bucket.key}>
         <Checkbox
-          label={label}
+          label={label ? label : bucket.key}
```
KonstantinaStoikou, line 141: Is this check necessary since the label is defined above and it will always have a value?

zzacharo, line 141: I think this check is redundant as you check the label before no?

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -75,13 +75,22 @@ class BucketAggregationValues extends Component {
...
+        );
+      }
       return (
         <ValueElement
           key={bucket.key}
           bucket={bucket}
           isSelected={isSelected}
           onFilterClicked={onFilterClicked}
           getChildAggCmps={getChildAggCmps}
+          valueLabel={label}
```
zzacharo, line 93: minor: It is definitely not to be addressed now...but why don't we just name the prop `label`? The `value` prefix is unnecessary as the property refers to the label of the `ValueElement` component. I might miss something though :P But, again, not to be addressed now just curiosity :)

src/lib/components/EmptyResults/EmptyResults.js
```diff
@@ -35,6 +36,7 @@ class EmptyResults extends Component {
         <Element
           queryString={queryString}
           resetQuery={this.resetQuery}
+          extraContent={extraContent}
```
zzacharo, line 39: Where is this used? It's a property specific to `Overridable`?

src/lib/components/BucketAggregation/BucketAggregationValues.js
kprzerwa: I didn't want to confuse it with the whole aggregation label (the one you see on top of the filter)

src/lib/components/EmptyResults/EmptyResults.js
kprzerwa: no, it is if you want to pass something additional to the EmptyResult, f.e. button to add a new instance or more information. It is more flexible this way than to override the whole component just for this.

src/lib/components/BucketAggregation/BucketAggregationValues.js
kprzerwa: 👍

kprzerwa: 👍

src/lib/components/EmptyResults/EmptyResults.js
zzacharo: But is not used in the `Element`, no?

kprzerwa: no it's not used, but I think it should be passed as a prop

zzacharo: Ok I missed the fact that the Element is overridable :) Then, it looks good!

PR #107, opened 2020-04-14
package.json: add watch/link dist folder

zzacharo:
Part of proposed solution https://github.com/inveniosoftware/invenio-app-rdm/issues/65#issuecomment-613406058

PR #133, opened 2020-08-19
searchApi: add hidden parameters option

### LINKED ISSUE #130: Add "hidden" filters/params
There are search page use cases that require a specific query string parameter to be passed to the underlying REST API request, but at the same time not to be serialized/updated in the URL.

A quick example is having a "Pending members" search page at `/members/pending`, that hits `/api/members?status=pending`, and allows other standard filters/queries. The top-bar URL doesn't really need the `status=pending` parameter to be displayed, but it has to be kept somehow in the query state.

To properly facilitate this feature, we would need to add a `query.hidden_filters` key in the store (along with the appropriate reducers/actions/etc.), which can be also initialized accordingly via config. This value can then be used in the underlying search API class to be injected into the request.

Glignos:
* Closes #130
[ci skip]

### DIFF
--- a/src/lib/api/contrib/invenio/InvenioRequestSerializer.js
+++ b/src/lib/api/contrib/invenio/InvenioRequestSerializer.js
@@ -8,6 +8,7 @@
 
 import Qs from 'qs';
 import _extend from 'lodash/extend';
+import _isEmpty from 'lodash/isEmpty';
 
 /** Default backend request serializer */
 export class InvenioRequestSerializer {
@@ -63,9 +64,9 @@ export class InvenioRequestSerializer {
    * @param {object} stateQuery the `query` state to serialize
    */
   serialize(stateQuery) {
-    const { queryString, sortBy, sortOrder, page, size, filters } = stateQuery;
+    const { queryString, sortBy, sortOrder, page, size, filters, hiddenParams } = stateQuery;
 
-    const getParams = {};
+    let getParams = {};
     if (queryString !== null) {
       getParams['q'] = queryString;
     }
@@ -81,6 +82,9 @@ export class InvenioRequestSerializer {
     }
     if (size > 0) {
       getParams['size'] = size;
+    }
+    if (!_isEmpty(hiddenParams)) {
+      _extend(getParams, this._addFilters(hiddenParams));
     }
     const filterParams = this._addFilters(filters);
     _extend(getParams, filterParams);
--- a/src/lib/store.js
+++ b/src/lib/store.js
@@ -14,7 +14,10 @@ import rootReducer from './state/reducers';
 import { INITIAL_STORE_STATE } from './storeConfig';
 
 export function configureStore(appConfig) {
-  const initialQueryState = INITIAL_STORE_STATE;
+  const initialQueryState = {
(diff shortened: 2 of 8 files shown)


src/lib/api/contrib/invenio/InvenioRequestSerializer.js
```diff
@@ -82,6 +82,9 @@ export class InvenioRequestSerializer {
     if (size > 0) {
       getParams['size'] = size;
     }
+    if (hidden_params !== null) {
```
ntarocco, line 85: very minor:
```suggestion
    if (!_isEmpty(hidden_params)) {
```

src/lib/store.js
```diff
@@ -26,10 +26,13 @@ export function configureStore(appConfig) {
     error: {},
   };
 
+
+
   // configure the initial state
   const preloadedQueryState = appConfig.urlHandlerApi
     ? appConfig.urlHandlerApi.get(initialQueryState)
     : initialQueryState;
+  preloadedQueryState.hidden_params = appConfig.searchApi.axiosConfig.hidden_params
```
ntarocco, line 35: While the idea was to keep the `state` as clean as possible and to store all user choices/input only, I like the simplicity of your solution. 👏 
My fear, and this is what we need to keep an eye on, is that we don't abuse of this. If we see the need to even add more things similar to this one, then we need to find a better solution.

I would really stress on remembering (maybe create an issue) on removing this when `axios` gets fixed, and probably add a comment here and/or in the state below to explain why.

ntarocco (approved):

src/lib/store.js
slint: I would say that even after the `axios` issue gets fixed, it would still be good to store the hidden params in the state, i.e. making this not just a way to configure the search API client, but also a capability of RSK in general.

Also, another idea for this feature is to add this fully to the state since you might want to change these hidden params even after the app's initialization, e.g. you have in the UI tab views for "pending", "accepted", "rejected" and when the user clicks them, you change the hidden params state (`status=pending`), trigger a reload of the results, but don't update the browser URL with different `f=...` params.

lnielsen:
@ntarocco @slint - Just checking - is this PR good to get merged and released? What new issues need to be created? 
Is it just one issue about reviewing state management in RSK?

src/lib/api/contrib/invenio/InvenioRequestSerializer.js
```diff
@@ -82,6 +83,9 @@ export class InvenioRequestSerializer {
     if (size > 0) {
       getParams['size'] = size;
     }
+    if (!_isEmpty(hiddenParams)) {
+      getParams = {...getParams, ...hiddenParams};
```
slint, line 87: I think `hiddenParams` has to pass through `this._addFilters(hiddenParams);`, to get translated to proper querystring parameters.

slint (commented):
One minor issue on how the hidden params get serialized for the request, otherwise LGTM to merge

slint:
Includes #107

slint:
@ntarocco if this one looks good to merge, to release is the process the same as for Invenio modules, i.e. make release commit and push a tag?

PR #137, opened 2020-10-05
setInitialState: set state only in ReactSearchKit component

### LINKED ISSUE #110: The `setInitialState` in each component runs after the first search query
Each component has the method `setInitialState` to set the initial values to be used in the first search query.
The issue is that this runs after the first search query, so it is not take into account.

Reproduced in `invenio-app-ils` with a SortBy component.

KonstantinaStoikou:
The query state is initialized only in the ReactSearchKit component because before the query for the components SortBy, Pagination etc. was executed before their mounting so the query parameters were not included.

Closes #110


ntarocco (commented):
It looks like that `defaultSortByOnEmptyQuery` is not used at all, probably another bug. It is important we fix it...
We should probably add to the query state:
```
sortBy: ...,
sortByOnEmptyQuery: ... || sortBy, (in case is not set)
```

Then, when executing the search, we should check that if the query string is empty, it should use `sortByOnEmptyQuery`, otherwise `sortBy`.
Would it be possible to add this too?

src/lib/components/Pagination/Pagination.js
```diff
@@ -90,7 +79,6 @@ Pagination.defaultProps = {
     showPrev: true,
     showNext: true,
   },
-  defaultValue: 10,
```
FlorianCassayre, line 93: :tada:
(this mere line was the source of our issue)

src/lib/components/ReactSearchKit/ReactSearchKit.js
```diff
@@ -18,17 +18,30 @@ import { buildUID } from '../../util';
...
+    const validate = (initialQueryState) => {
+      const layoutOptions = ['list', 'grid'];
+      if (
+        initialQueryState['layout'] &&
+        !layoutOptions.includes(initialQueryState['layout'])
+      ) {
+        throw new Error('Layout option is invalid.');
+      }
+
+      return initialQueryState;
```
ntarocco, line 31: Sorry, I was not extremely clear... :( it would be great also if you pass an extra key, it fails.
e.g. `initialQueryState={'hello': 'world', ...}` should fail because not a valid key for the state.

pseudocode:
```
for key in initalQueryState:
  if key not in INITIAL_STATE.keys():
       throw...
```

other validations maybe: page should be a number, anything else?

src/lib/components/ReactSearchKit/ReactSearchKit.test.js
```diff
@@ -124,15 +141,20 @@ describe('test ReactSearchKit component', () => {
     );
   });
 
-  it('should use inject defaultSortByOnEmptyQuery in the configuration', () => {
+  it('should use layout options in the initialQueryState', () => {
     shallow(
-      <ReactSearchKit searchApi={searchApi} defaultSortByOnEmptyQuery="value" />
+      <ReactSearchKit
```
ntarocco, line 146: nice thanks a lot!!!
Can you also add a test if you pass something else that is not list/grid?
And if you pass an invalid key?

ntarocco (commented):
thank you for the massive work, for the documentation too!

ntarocco (approved):
@FlorianCassayre Were you able to test in ILS to see if everything works?
LGTM! We will have to document the breaking changes in the release.

FlorianCassayre:
No, I didn't test in ILS because Konstantina's PR on ILS was already fixing the pagination bug. I'll quickly try now, check that it doesn't break anything.

FlorianCassayre:
Do not merge, something broke.

FlorianCassayre (approved):
Successfully tested on ILS, looks good to me

PR #141, opened 2020-10-08
components: SearchBar functions override

topless:
- Moved functionality from Element to the parent component SearchBar, so that we can supply them to the Element that overrides.

### DIFF
--- a/src/lib/components/SearchBar/SearchBar.js
+++ b/src/lib/components/SearchBar/SearchBar.js
@@ -31,70 +31,125 @@ class SearchBar extends Component {
     this.updateQueryString(this.state.currentValue);
   };
 
+  onBtnSearchClick = (event, input) => {
+    this.executeSearch();
+  };
+
+  onKeyPress = (event, input) => {
+    if (event.key === 'Enter') {
+      this.executeSearch();
+    }
+  };
+
   render() {
-    const { placeholder, overridableId } = this.props;
+    const {
+      actionProps,
+      autofocus,
+      executeSearch,
+      onBtnSearchClick,
+      onInputChange,
+      onKeyPress,
+      overridableId,
+      placeholder,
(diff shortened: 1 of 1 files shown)


src/lib/components/SearchBar/SearchBar.js
```diff
@@ -31,33 +31,54 @@ class SearchBar extends Component {
...
+    }
+  };
+
   render() {
     const { placeholder, overridableId } = this.props;
     return (
       <Element
-        placeholder={placeholder}
-        queryString={this.state.currentValue}
+        onBtnSearchClick={this.onBtnSearchClick}
```
ntarocco, line 48: I am bit lost here, shouldn't you use the prop?
```suggestion
        const { placeholder, overridableId, ..., onBtnSearchClick, ... } = this.props;
        onBtnSearchClick={onBtnSearchClick}
```
same for other props.

src/lib/components/SearchBar/SearchBar.js
```diff
@@ -31,33 +31,63 @@ class SearchBar extends Component {
     this.updateQueryString(this.state.currentValue);
   };
 
+  onBtnSearchClick = (event, input) => {
```
ntarocco, line 34: I am not sure that you have to make these 2 methods overridable via props. If it is because it is used in the homepage of ILS, then the cmp in the homepage should be removed and replaced with a simple fake search input.
Is it needed for other use cases?
If I remember correctly, you only need to make the "style" configurable via prop.

ntarocco (approved):

rollup.config.js
```diff
@@ -24,6 +24,8 @@ export default {
       sourcemap: true,
     },
   ],
+  context: 'null',
```
ntarocco, line 27: out of curiosity, was this needed?

PR #142, opened 2020-10-15
global: update query state from response

zzacharo:


src/lib/api/contrib/elasticsearch/ESResponseSerializer.js
```diff
@@ -16,10 +16,12 @@ export class ESResponseSerializer {
...
   serialize(payload) {
+    const { aggregations, hits, ...extras } = payload;
     return {
-      aggregations: payload.aggregations || {},
-      hits: payload.hits.hits.map(hit => hit._source),
-      total: payload.hits.total.value,
+      aggregations: aggregations || {},
+      hits: hits.hits.map((hit) => hit._source),
+      total: hits.total.value,
+      extras: extras,
```
ntarocco, line 24: this `extras` is really a black hole, an open door to anything :(

src/lib/components/ReactSearchKit/ReactSearchKit.js
```diff
@@ -26,6 +26,7 @@ export class ReactSearchKit extends Component {
         ? props.urlHandlerApi.customHandler ||
           new UrlHandlerApi(props.urlHandlerApi.overrideConfig)
         : null,
+      updateQueryStateFromResponse: props.updateQueryStateFromResponse,
```
ntarocco, line 29: IMHO, this should be handled by the response serializer only, so an option there and not here at the RSK app. WDYT?

ntarocco (commented):
Since you want to modify the `queryState` on response, why not:
* keep the `extras` only at the level of the Invenio response serializer. If you use another serializer, e.g. Elasticsearch, probably you will never use this
* change the queryState in the response serializer. It is not idea, but at least it does not "pollute" the entire app.

Let me know what you think :)

src/lib/components/ReactSearchKit/ReactSearchKit.js
zzacharo: Hmm but the serializer doesn't know how to update the state...I need to modify the way the `executeQuery` action is working...what did you have in mind?

src/lib/api/contrib/elasticsearch/ESResponseSerializer.js
zzacharo: This, I just did it for completeness to comply with the Invenio one...I am fine with omitting it :). The `extras` is not mandatory!

src/lib/api/contrib/invenio/InvenioSearchApi.js
```diff
@@ -78,9 +81,18 @@ export class InvenioSearchApi {
...
+    response = this.responseSerializer.serialize(response.data);
+    const newQueryState = updateQueryState(
+      stateQuery,
+      response.extras,
+      STORE_KEYS
+    );
+    if (!_isEmpty(newQueryState)) {
+      response.newQueryState = newQueryState;
+    }
+    return response;
```
ntarocco, line 96: delete the `extras` before? (double check if it works if not set!)
```suggestion
    delete response.extras
    return response;
```

src/lib/state/selectors/query.js
```diff
@@ -120,14 +120,14 @@ export const updateQueryFilters = (queryFilter, stateFilters) => {
...
-    filter => (tempStateFilters = updateFilter(filter, tempStateFilters))
+    (filter) => (tempStateFilters = updateFilter(filter, tempStateFilters))
   );
   return tempStateFilters;
 };
 
 export const updateQueryState = (oldState, newState, storeKeys) => {
   let pickedState = _pick(newState, storeKeys);
-  if ('filters' in pickedState) {
+  if ('filters' in pickedState && !_isEmpty(pickedState.filters)) {
```
ntarocco, line 130: maybe the opposite?
```suggestion
  if (!_isEmpty(pickedState.filters) && 'filters' in pickedState) {
```

ntarocco (commented):
I prefer this version! :)

ntarocco (approved):

PR #143, opened 2020-10-16
buckets: passes provided labels to children

Glignos:
* Addresses https://github.com/inveniosoftware/invenio-rdm-records/issues/202

### DIFF
--- a/src/lib/components/BucketAggregation/BucketAggregationValues.js
+++ b/src/lib/components/BucketAggregation/BucketAggregationValues.js
@@ -31,7 +31,7 @@ class BucketAggregationValues extends Component {
     );
   };
 
-  getChildAggCmps = (bucket, selectedFilters) => {
+  getChildAggCmps = (bucket, selectedFilters, valuesLabels) => {
     const hasChildAggregation =
       this.childAgg && this.childAgg['aggName'] in bucket;
     let selectedChildFilters = [];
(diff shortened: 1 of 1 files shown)


src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -129,7 +130,8 @@ const ValueElement = (props) => {
     overridableId,
   } = props;
 
-  const label = valueLabel ? valueLabel : `${bucket.key} (${bucket.doc_count})`;
+  const label = valueLabel ? `${valueLabel} (${bucket.doc_count}) `:
```
zzacharo, line 133: I think that the `valueLabel` should be able to override the whole value i.e not force the `doc_count` to appear! I would say that we can go with 2 solutions:
1) Compute the `${valueLabel} (${bucket.doc_count}) ` beforehand when we build the value map or
2) Create another prop e.g 
```javascript
const templateLabel = (valueLabel, bucket) => valueLabel ? `${valueLabel} (${bucket.doc_count})` : `${bucket.key} (${bucket.doc_count})`
```
so then the above line would be something like
```javascript
const label = props.templateLabel(valueLabel, bucket);
```

In this way, you have property that can construct your final template based on the bucket! Your `valueLabels` map can remain still plain values! @ntarocco WDYT? 

**Context**: We need this to override the text of nested aggregations in RDM. See https://github.com/inveniosoftware/invenio-rdm-records/issues/202

ntarocco (commented):
I have commented on the issue already. Honestly I really think we should **NOT** do this: labels can be thousands, potentially millions and it is one point to be addressed in the vocabularies modules first.

I would suggest a discussion/brainstorming on this, I don't see how this can help.

PR #144, opened 2020-10-20
aggs: remove valueLabels property

zzacharo:
* You can override the `agg.bucket` label by injecting a `label` key to
  the `agg.buckets` items

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+Version 1.0.0-alpha.7 (released 2020-10-20)
+
+* Override query state from search response.
+* `InvenioSearchApi` extracts overriden state from response.
+* Configure `Sort` component to disable the order option.
+* Remove `valuesLabels` prop from `BucketAggregation` component.
+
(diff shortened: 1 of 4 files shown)


ntarocco (approved):
LGTM! I saw/heard that you were opting for a kind of a template, maybe in the future it could be a good idea if you want to do something like `<label (count)>`

lnielsen:
Shouldn't the count be in a separate element somehow so we can do these round circles around the numbers:
Screenshot 2020-10-20 at 15 55 04

zzacharo:
@ntarocco @lnielsen if you want I can amend this PR with a `template` prop, so what @lnielsen show can be feasible!

zzacharo:
@lnielsen For the count, we will override this in `invenio-app-rdm` similar to https://github.com/inveniosoftware/react-invenio-app-ils/blob/75f1faf53b973c2a44d72905115513ffc250730b/src/lib/modules/SearchControls/overridden/CardBucketAggregationValueElement.js#L5

PR #145, opened 2020-10-21
components: adds toggle filter component

Glignos:


src/lib/components/Toggle/index.js
```diff
@@ -0,0 +1,22 @@
...
+import { updateQueryFilters } from '../../state/actions';
+import ToggleComponent from './ToggleElement';
+
+const mapDispatchToProps = dispatch => ({
+  updateQueryFilters: filter => dispatch(updateQueryFilters(filter)),
+});
+
+export const Toggle = connect(
+  state => ({
+    userSelectionFilters: state.query.filters
```
zzacharo, line 19: Not used :)

src/lib/components/Toggle/ToggleElement.js
```diff
@@ -0,0 +1,49 @@
...
+        this.updateQueryFilters = props.updateQueryFilters;
+      }
+
+      onToggleClicked = () => {
+        this.updateQueryFilters(this.filterValue);
+      };
+      
+      render() {
+        return(
+        <Overridable id={'ToggleComponent'}>
```
zzacharo, line 31: Are you sure the overridable id is ok to be the same as the name of the component [here](https://github.com/inveniosoftware/react-searchkit/pull/145/files#diff-1afdea65744f0f08c5e62dbe48dbc8eaafd928c83968f554d4c0aca3e862bdcfR49) ? Also, we will need to pass down the props i.e `<Overridable id={'ToggleComponent'}> {...this.props}` so they can be available when you override the component I think....

src/lib/components/Toggle/Toggle.js
```diff
@@ -0,0 +1,79 @@
...
+        super(props);
+        this.title = props.title;
+        this.label = props.label;
+        this.filterValue = props.filterValue;
+        this.updateQueryFilters = props.updateQueryFilters;
+        this.overridableId = props.overridableId;
+      }
+
+      _isChecked = (userSelectionFilters) => {
+        let isFilterActive = !_isEmpty(userSelectionFilters.filter(
```
ntarocco, line 28:
```suggestion
        const isFilterActive = userSelectionFilters.filter(...).length > 0
```

src/lib/components/Toggle/Toggle.js
```diff
@@ -0,0 +1,79 @@
...
+import Overridable from 'react-overridable';
+import _get from 'lodash/get';
+import { Card } from 'semantic-ui-react';
+import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
+
+class ToggleComponent extends Component{
+    constructor(props) {
+        super(props);
+        this.title = props.title;
```
ntarocco, line 20: very minor: because of a "bad" habit, we started doing this in the constructor... it is probably not needed, and props can be used directly. Feel free to improve code! :)

src/lib/components/Toggle/index.js
ntarocco: Not sure what you mean, but it is used right?

ntarocco (approved):
LGTM!
Minor: it would be great if you create a new doc page, even empty, only title, so that at some point we document this component too and it is not forgotten!

src/lib/components/Toggle/index.js
zzacharo: Yes it was updated I think...Eitherway, I will resolve it!

PR #149, opened 2020-11-01
RSK: implement sorting on empty query string

### LINKED ISSUE #111: Change sort by/order on empty query
The sort by/order on empty query has been removed because it was not working.

## Requirements

The user should be able to set the initial state of the app and also define what is the sort by/order when the query string is empty.
Example:
- when the query string is not empty, the order should be the initial (e.g. `bestmatch`)
- when the query string is empty, the order should be the passed one (e.g. `latest`)

If the user changes and select a specific sort by/order, then her/his selection takes precedence and it is user for all queries.

On top of that, the SortOrder component was not taking into account the `defaultOnEmptyQuery` options.
As consequence, when searching with a query or with empty query, the SortBy component changes the default value but not the SortOrder.

For example, when searching with query, the search should be probably `bestmatch:asc`, while when empty `mostrecent:desc` (newest first).

ntarocco:
* introduces a new prop to define which sorting to use on empty query
  string
* adds a new private app state to the store to keep track if the user
  has selected a new sorting and what was the initial sorting before
  change
* adds tests
* changes documentation
* closes #111

### DIFF
--- a/docs/docs/components/react_search_kit.md
+++ b/docs/docs/components/react_search_kit.md
@@ -38,14 +38,34 @@ See the [complete guide](main_concepts.md) for detailed information.
 
   - **customHandler** `object`: override entirely the default class `UrlHandlerApi`.
 
-- **searchOnInit** `object` _optional_
+- **searchOnInit** `boolean` _optional_
 
   A boolean to perform a search when the application is mounted. Default `true`.
 
 - **appName** `string` _optional_
 
-  A name identifier to distinguish uniquely the application. Default `RSK`.
+  A name identifier to distinguish uniquely the application. Useful if multiple ReactSearchKit apps are loaded in the same page. Default `RSK`.
 
 - **eventListenerEnabled** `boolean` _optional_
 
   If `true` the application listens to the `queryChanged` event else if `false` no listener is registered. When this event is emitted the application triggers a search based on the payload that is passed to the event at the emission time. Default `false`.
+
+- **initialQueryState** `object` _optional_
+
+  Set the initial state of your ReactSearchKit application. It will be used to render each component with these default selected values and to perform the first search query, if `searchOnInit` is set to `true`.
+  The object keys must match the query state fields and the values must be valid values that correspond to the values passed to each parameters.
+
+- **defaultSortingOnEmptyQueryString** `object` _optional_
+
+  It is sometimes useful to automatically change the default sorting in case the query string is set or left empty.
+  A typical case is when your app should return by default the most recent items when the user did not search for anything in particular, but it should instead return the best matching items when searching with a particular query string.
+  When enabled, the behavior will be the following:
+
+    * if the user did **not** change sorting, the default sorting with an **empty** query string will be the one defined with this prop `defaultSortingOnEmptyQueryString`.
+    * if the user did **not** change sorting, the default sorting with a query string will be the one defined with the prop `initialQueryState`.
+    * if the user **did** change sorting and selected a value, then the user preference will be kept, independently of the presence or absence of the query string.
+
+  - **sortBy** `string`: the query state `sortBy` value to use on empty query string.
+
+  - **sortOrder** `string`: the query state `sortOrder` value to use on empty query string.
+
--- a/src/lib/components/Toggle/Toggle.js
+++ b/src/lib/components/Toggle/Toggle.js
@@ -6,20 +6,18 @@
  * under the terms of the MIT License; see LICENSE file for more details.
  */
 
-import React, { Component } from 'react';
-import { Checkbox } from 'semantic-ui-react';
-import Overridable from 'react-overridable';
-import _get from 'lodash/get';
-import { Card } from 'semantic-ui-react';
 import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import Overridable from 'react-overridable';
+import { Card, Checkbox } from 'semantic-ui-react';
 
 class ToggleComponent extends Component {
-
   _isChecked = (userSelectionFilters) => {
-    const isFilterActive = userSelectionFilters.filter(
-      (filter) => filter[0] === this.props.filterValue[0]
-    ).length > 0
-    return isFilterActive
+    const isFilterActive =
+      userSelectionFilters.filter(
+        (filter) => filter[0] === this.props.filterValue[0]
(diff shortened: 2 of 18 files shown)


docs/docs/components/react_search_kit.md
```diff
@@ -38,14 +38,30 @@ See the [complete guide](main_concepts.md) for detailed information.
...
   If `true` the application listens to the `queryChanged` event else if `false` no listener is registered. When this event is emitted the application triggers a search based on the payload that is passed to the event at the emission time. Default `false`.
+
+- **initialQueryState** `object` _optional_
+
+  Set the initial state of your ReactSearchKit application. It will be used to render each component with these default selected values and to perform the first search query, if `searchOnInit` is set to `true`.
+  The object keys must match the query state fields and the values must be valid values that correspond to the values passed to each parameters.
+
+- **defaultSortingOnEmptyQueryString** `object` _optional_
+
+  It is sometimes useful to automatically change the default sorting in case the query string is set or empty.
```
zzacharo, line 60:
```suggestion
  It is sometimes useful to automatically change the default sorting in case the query string is set or left empty.
```

docs/docs/components/react_search_kit.md
```diff
@@ -38,14 +38,30 @@ See the [complete guide](main_concepts.md) for detailed information.
...
+  A typical case is when your app should return by default the most recent items when the user did not search for anything in particular, but it should instead return the best matching items when searching with a particular query string.
+  You can enabled this behavior by defining in the `initialQueryState` the expected sorting when the query string is set by the user, and an alternative sorting when empty.
```
zzacharo, line 62:
```suggestion
  You can enable this behavior by defining in the `initialQueryState` the expected sorting when the query string is set by the user, and an alternative sorting when left empty.
```

src/lib/components/Toggle/Toggle.js
```diff
@@ -6,42 +6,36 @@
...
-      overridableId,
-      title,
-      label
-    } = this.props;
+    const { userSelectionFilters, overridableId, title, label } = this.props;
     var isChecked = this._isChecked(userSelectionFilters);
-    const onToggleClicked = this.onToggleClicked
-    debugger
+    const onToggleClicked = this.onToggleClicked;
+    debugger;
```
zzacharo, line 31: leftover? Was this merged before ????

src/lib/state/actions/query.js
```diff
@@ -137,37 +148,104 @@ export const resetQuery = () => {
...
   shouldUpdateUrlQueryString = true,
   shouldReplaceUrlQueryString = false,
 } = {}) => {
   return async (dispatch, getState, config) => {
-    let queryState = _cloneDeep(getState().query);
+    const appState = getState().app;
+    let queryState = getState().query;
+    updateQueryStateSorting(queryState, appState, config);
+
+    queryState = _cloneDeep(queryState);
```
zzacharo, line 234: Maybe better to return a cloned state from the `updateQueryStateSorting `? Then, if someone uses it in another action, there wouldn't need to care about the mutation that is happening inside the function...

src/lib/state/actions/query.js
```diff
@@ -137,37 +148,104 @@ export const resetQuery = () => {
...
+};
+
+/**
+ * Update query state and URL args with the new query state given by the backend response.
+ * @param {object} response - API response
+ * @param {func} dispatch - Redux `dispath` function
+ * @param {func} getState - function to get the Redux state
+ * @param {object} appConfig - app config
+ */
+const updateQueryStateAfterResponse = (
```
zzacharo, line 181: I am a bit hesitant because now this file includes redux action + helper functions...maybe prefix them with `_` or we have a helpers file? WDYT?

ntarocco: Honestly I would refactor this file, to create objects that encapsulate private stuff e.g. `updateQueryString obj and executeSearch obj`... but it requires more large changes. What about if I create a new PR with such refactoring?

src/lib/components/Toggle/Toggle.js
ntarocco: oh wow it looks like yes...are you opening the Developer Tools when working on RDM? ping @Glignos :) :D

Glignos: @ntarocco Oh well, that's on me. :)

KonstantinaStoikou (commented):
I ran the branch locally and noticed that when you enter a query and then change the sorting and clear the query the last sorting is used in the empty query. Is this the desired behavior or everytime an empty query is entered the sorting should change to the `defaultOnEmptyString`?

src/lib/state/actions/query.js
zzacharo: Yes even opening just an issue is fine with me :)

zzacharo:
> I ran the branch locally and noticed that when you enter a query and then change the sorting and clear the query the last sorting is used in the empty query. Is this the desired behavior or everytime an empty query is entered the sorting should change to the `defaultOnEmptyString`?

I would say that everytime the query string is empty the `defaultOnEmptyString` should be used :)

src/lib/components/Toggle/Toggle.js
ntarocco: fixed

src/lib/state/actions/query.js
ntarocco: I have changed, can you double check? I am not sure I fully understood your comment.

ntarocco:
> > I ran the branch locally and noticed that when you enter a query and then change the sorting and clear the query the last sorting is used in the empty query. Is this the desired behavior or everytime an empty query is entered the sorting should change to the `defaultOnEmptyString`?
> 
> I would say that everytime the query string is empty the `defaultOnEmptyString` should be used :)

Good question @KonstantinaStoikou @zzacharo how I have implemented is:
- nothing touched, no query string: use sorting given via config `onEmptyQuery`
- nothing touched, with query string: use default sorting
- user changed sorting (user choice), independently of the presence of the query string: use user choice

The idea was that if the user selects a sorting, it means (s)he wants that. If I choose `sort by most loaned`, I am not sure I wanted to be automatically changed back to most recent or bestmatch. WDYT?

zzacharo (approved):

PR #157, opened 2020-11-09
fix url not updating state

FlorianCassayre:
* fixes url history not updating the internal state
* closes inveniosoftware/react-invenio-app-ils#274

Tested on zenodo demo and on ILS. Also tested on CDS videos demo which has disabled support for url handling.

It may not be the best way to do it, but it works. I'm preventing the url from being updated twice: this is a workaround (I hope temporary) because the url is replaced twice, and even though the state is equivalent in both cases the order of the attributes is not.

Open to suggestions.

### DIFF
--- a/src/lib/state/reducers/query.js
+++ b/src/lib/state/reducers/query.js
@@ -6,7 +6,7 @@
  * under the terms of the MIT License; see LICENSE file for more details.
  */
 
-import { INITIAL_QUERY_STATE_KEYS } from '../../storeConfig';
+import { INITIAL_QUERY_STATE, INITIAL_QUERY_STATE_KEYS } from '../../storeConfig';
 import { updateQueryFilters, updateQueryState } from '../selectors';
 import {
   CLEAR_QUERY_SUGGESTIONS,
@@ -92,7 +92,8 @@ export default (state = {}, action) => {
     case SET_QUERY_STATE:
       return {
         ...state,
-        ...updateQueryState(state, action.payload, INITIAL_QUERY_STATE_KEYS),
+        ...INITIAL_QUERY_STATE,
+        ...updateQueryState(INITIAL_QUERY_STATE, action.payload, INITIAL_QUERY_STATE_KEYS),
       };
     case RESULTS_UPDATE_LAYOUT:
       return {
--- a/src/lib/state/reducers/results.js
+++ b/src/lib/state/reducers/results.js
@@ -13,6 +13,11 @@ import {
 } from '../types';
 
 export default (state = {}, action) => {
+  // A completion event (success or error) should only update the state if it was
+  // originally triggered by the latest request made up to that point.
+  // Otherwise the result shall be ignored because there is a more recent request
+  // either waiting to complete, or that has already completed.
+  const isLatestRequest = action.payload && state.queryState === action.payload.queryState;
   switch (action.type) {
     case RESULTS_LOADING:
       return {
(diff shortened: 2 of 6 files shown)


FlorianCassayre:
I used the following widespread design for requests:
* any request that is made triggers a loading state
* a completion (which can be successful or not) is accepted only if the corresponding request is the latest that was made, otherwise it is ignored

src/lib/state/reducers/results.js
```diff
@@ -21,29 +21,38 @@ export default (state = {}, action) => {
...
-        loading: false,
-        data: {
-          ...state.data,
-          aggregations: {},
-          hits: [],
-          total: 0,
-        },
-        error: action.payload,
-      };
+      if (state.queryState === action.payload.queryState) {
```
zzacharo, line 42: Maybe this can be done in the action level so that the reducer function stays pure :)

src/lib/state/reducers/results.js
```diff
@@ -21,29 +21,38 @@ export default (state = {}, action) => {
...
-        loading: false,
-        data: {
-          ...state.data,
-          aggregations: action.payload.aggregations,
-          hits: action.payload.hits,
-          total: action.payload.total,
-        },
-        error: {},
-      };
+      if (state.queryState === action.payload.queryState) {
```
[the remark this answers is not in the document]
FlorianCassayre: This is was I tried to explain in the above comment, namely there can be several concurrent requests running at the same time (for instance if the user goes back two pages, relatively quickly). In that case we only want to show trigger the completion event when the last request terminates, and ignore (or cancel) the others.
It's possible to use a counter for that purpose but a reference works too.

src/lib/state/reducers/results.js
FlorianCassayre: I'm not sure if it will work because we need to know the current state in order to update it accordingly.

ntarocco (commented):
Thanks a lot for your explanation. If leave aside the race condition for a moment, I am not sure I understand what is causing the browser back issue.
It would be great if tomorrow we discuss this IRL! :)

FlorianCassayre:
PR was revised:
* the action `SET_QUERY_STATE` was _updating_ the state instead of _setting_ (= replacing) it, I fixed it and removed the newly created action
* the parameters of the url are now sorted lexicographically by their key, except for `q` which is meant to appear first
* removed the temporary parameter `replaceUrl` since now the replaced url has no change

src/lib/state/reducers/query.js
```diff
@@ -92,7 +92,8 @@ export default (state = {}, action) => {
     case SET_QUERY_STATE:
       return {
         ...state,
-        ...updateQueryState(state, action.payload, INITIAL_QUERY_STATE_KEYS),
+        ...INITIAL_QUERY_STATE,
```
ntarocco, line 95: As discussed, we should keep it to avoid that extra keys, coming externally (backend response), will be injected in the query state.

I would also rename the method `updateQueryState` to something more clear, e.g. `sanitize...`

src/lib/state/reducers/results.js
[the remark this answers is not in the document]
ntarocco: I would suggest to make this explicit in the code, e.g.:
```
// comment explaining what is happening
isLatestRequest = state.queryState === action.payload.queryState
```

src/lib/state/reducers/results.js
ntarocco: same as above

FlorianCassayre:
Changes:
```diff
    case SET_QUERY_STATE:
      return {
        ...state,
        ...INITIAL_QUERY_STATE,
-       ...action.payload,
+       ...updateQueryState(INITIAL_QUERY_STATE, action.payload, INITIAL_QUERY_STATE_KEYS),
      };
```

`INITIAL_QUERY_STATE` is required for the filters to be correctly reset.

ntarocco (approved):

PR #161, opened 2020-11-10
handle malformed urls

FlorianCassayre:
* addresses inveniosoftware/react-invenio-app-ils#128
* also exposes the class `UrlParamValidator` in case one needs to build a stricter validator on top of it (for instance, ensure the provided page size is indeed an option of the dropdown)

### DIFF
--- a/src/lib/api/UrlParamValidator.js
+++ b/src/lib/api/UrlParamValidator.js
@@ -1,0 +1,45 @@
+/*
+ * This file is part of React-SearchKit.
+ * Copyright (C) 2020 CERN.
+ *
+ * React-SearchKit is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import _isArray from 'lodash/isArray';
+import _isNumber from 'lodash/isNumber';
+
+/** Default implementation for a param validator class */
+export class UrlParamValidator {
+  /**
+   * Return true if the param value is valid, false otherwise
+   * @param {object} urlHandler the url handler
+   * @param {string} key the parameter key
+   * @param {string} value the parameter value
+   */
+  isValid = (urlHandler, key, value) => {
+    switch (key) {
+      case 'queryString':
+      case 'sortBy':
+        return true;
+      case 'sortOrder':
+        return ['asc', 'desc'].includes(value);
+      case 'page':
+      case 'size':
+        return _isNumber(value) && value > 0;
+      case 'layout':
+        return ['grid', 'list'].includes(value);
+      case 'filters':
+      case 'hiddenParams':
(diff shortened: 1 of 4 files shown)


src/lib/api/UrlParamValidator.js
```diff
@@ -0,0 +1,42 @@
...
+      case 'sortOrder':
+        return ['asc', 'desc'].includes(value);
+      case 'page': case 'size':
+        return _isNumber(value) && value > 0;
+      case 'layout':
+        return ['grid', 'list'].includes(value);
+      case 'filters': case 'hiddenParams':
+        const array = _isArray(value) ? value : [value];
+        const cKvSep = ':', cChildSep = urlHandler.urlFilterSeparator;
+        const pLiteral = `[^\\${cKvSep}\\${cChildSep}]*`,
```
ntarocco, line 33: oh wow, regex pro here! :D 👏

src/lib/api/UrlParamValidator.js
```diff
@@ -0,0 +1,42 @@
...
+export class UrlParamValidator {
+  /**
+   * Return true if the param value is valid, false otherwise
+   * @param {object} urlHandler the url handler
+   * @param {string} key the parameter key
+   * @param {string} value the parameter value
+   */
+  isValid = (urlHandler, key, value) => {
+    switch (key) {
+      case 'queryString': case 'sortBy':
```
ntarocco, line 22: funny formatting... I couldn't see the `sortBy`!

ntarocco (approved):
Just to understand, this is to validate the value of each param?
When we created, I think we thought that if values are wrong, it is because of URL manipulation and then it is out of RSK context to handle this.
But since you did, good job! :)

FlorianCassayre:
I see, I was not 100% sure about this part but it can be changed for sure (raise an exception instead of filtering the invalid parameters). Or even better, have it as a configurable behaviour. One argument in favour of filtering was to be able to gracefully handle old urls after one of the parameter name has changed; or for instance the allowed page sizes are updated. 
In any case the existing validation was incomplete and would let invalid values pass through, which we don't want for sure.

src/lib/api/UrlParamValidator.js
FlorianCassayre: (a bit verbose but I can make i a for-loop if too confusing)

src/lib/api/UrlParamValidator.js
FlorianCassayre: Thanks, will add a line-break.

FlorianCassayre:
Post-discussion: yes filter them, no need for config.

ntarocco (approved):

PR #164, opened 2020-11-20
fix sort state

FlorianCassayre:
* fixes the sorting filter not behaving properly since #149, namely the two following issues were resolved:
  * if the query is empty, then the sorting state was correctly set to `defaultSortingOnEmptyQueryString` but the url was not updated accordingly
  * if the page is reloaded, then the sorting parameters in the url were not used and the state would use the default value instead
* fixes a non-related programming mistake (the intent was assumed)

Tested on ILS, also made sure that it didn't break the history feature. Now:
* if the user did not change the sortby value *and* the url it came from did not contain a `sort` parameter, then:
  * the sortby value both in the state and in the url will depend on the query being empty or not, and will update accordingly so long as the user does not manually change the sortby value
* otherwise the sortby value will follow the user's choice

### DIFF
--- a/src/lib/components/Sort/Sort.js
+++ b/src/lib/components/Sort/Sort.js
@@ -52,7 +52,6 @@ class Sort extends Component {
     return (
       <ShouldRender
         condition={
-          currentSortBy !== null &&
           (sortOrderDisabled || currentSortBy !== null) &&
           !loading &&
           totalResults > 0
--- a/src/lib/api/UrlHandlerApi.js
+++ b/src/lib/api/UrlHandlerApi.js
@@ -220,6 +220,9 @@ export class UrlHandlerApi {
           result[queryStateKey] = urlParamsObj[paramKey].map((filter) =>
             this._filterStringToList(filter)
           );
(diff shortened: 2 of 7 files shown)


KonstantinaStoikou (approved):

kprzerwa (approved):

src/lib/components/Sort/Sort.js
```diff
@@ -52,7 +52,6 @@ class Sort extends Component {
     return (
       <ShouldRender
         condition={
-          currentSortBy !== null &&
```
FlorianCassayre, line 55: In particular please review this @zzacharo.

FlorianCassayre:
The flag "the user has changed sorting" now behaves as a computed property, depending on whether the sorting parameters appear in the url or not. Previously there was one side-effect: if the user reloaded the page then it would be assumed that the sorting had changed (because the parameter was in the url).

I left the other commit as a reference, but feel free to fixup or ask me to do it.

ntarocco:
Closed in favor of: https://github.com/inveniosoftware/react-searchkit/pull/166

PR #172, opened 2021-01-07
aggs: Allows objects to be passed as aggs

jrcastro2:
* closes https://github.com/inveniosoftware/react-invenio-app-ils/issues/22

### DIFF
--- a/src/lib/components/BucketAggregation/BucketAggregation.js
+++ b/src/lib/components/BucketAggregation/BucketAggregation.js
@@ -51,7 +51,17 @@ class BucketAggregation extends Component {
   _getResultBuckets = (resultsAggregations) => {
     // get buckets of this field
     const thisAggs = _get(resultsAggregations, this.agg.aggName, {});
-    return 'buckets' in thisAggs ? thisAggs['buckets'] : [];
+    if ('buckets' in thisAggs) {
+      if (!Array.isArray(thisAggs['buckets'])) {
+        // buckets can be objects or arrays: convert to array if object
+        // to keep it consistent
+        thisAggs['buckets'] = Object.entries(
+          thisAggs['buckets']
+        ).map(([key, value]) => ({ ...value, key }));
+      }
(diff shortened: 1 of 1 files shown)


src/lib/components/BucketAggregation/BucketAggregation.js
```diff
@@ -51,6 +51,14 @@ class BucketAggregation extends Component {
   _getResultBuckets = (resultsAggregations) => {
     // get buckets of this field
     const thisAggs = _get(resultsAggregations, this.agg.aggName, {});
+    if (thisAggs['buckets'] && !Array.isArray(thisAggs['buckets'])) {
+      let bucketArray = [];
+      Object.entries(thisAggs['buckets']).forEach(([key, value]) => {
+        value.key = key;
+        bucketArray.push(value);
+      });
+      thisAggs['buckets'] = bucketArray;
```
FlorianCassayre, line 60:
```suggestion
      thisAggs['buckets'] = Object.entries(thisAggs['buckets']).forEach(([key, value]) => ({ ...value, key }));
```
(minor suggestion)

FlorianCassayre (approved):

src/lib/components/BucketAggregation/BucketAggregation.js
```diff
@@ -51,6 +51,14 @@ class BucketAggregation extends Component {
   _getResultBuckets = (resultsAggregations) => {
     // get buckets of this field
     const thisAggs = _get(resultsAggregations, this.agg.aggName, {});
+    if (thisAggs['buckets'] && !Array.isArray(thisAggs['buckets'])) {
```
ntarocco, line 54: I would also a quick comment inside the if to explain: `// buckets can be objects or arrays: convert to array if object to keep it consistent`
I would also change the last `return` to avoid the double if:

```
if (...) {
    return thisAggs['buckets']
}
return [];

ntarocco (approved):

Issue #61, opened 2019-06-26
i18n: add support for translations


ntarocco:
As in the title. Align with Invenio chosen technology.

PR #173, opened 2021-01-18
count: format toLocaleString

### LINKED ISSUE #169: Count: number format
The counts components should format their numbers to improve readability. Ideally it should provide support for internationalization; otherwise we should stick to one locale for consistency (say `en_US`).
For instance in the following case:

image

...it should instead be displayed as:

image

There is another "count" component in the aggregation list, and which should also benefit from this feature:

image

mb-wali:
Applied for Count & BucketAggregation
closes #169

### DIFF
--- a/src/lib/components/BucketAggregation/BucketAggregationValues.js
+++ b/src/lib/components/BucketAggregation/BucketAggregationValues.js
@@ -123,7 +123,7 @@ const ValueElement = (props) => {
   } = props;
   const label = bucket.label
     ? bucket.label
-    : `${keyField} (${bucket.doc_count})`;
+    : `${keyField} (${bucket.doc_count.toLocaleString('en-US')})`;
   const childAggCmps = getChildAggCmps(bucket);
   return (
     <Overridable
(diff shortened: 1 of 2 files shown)


src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -123,7 +123,7 @@ const ValueElement = (props) => {
   } = props;
   const label = bucket.label
     ? bucket.label
-    : `${keyField} (${bucket.doc_count})`;
+    : `${keyField} (${bucket.doc_count.toLocaleString('en-US')})`;
```
zzacharo, line 126: I guess is fine for now to stick to `en-Us` locale but we should definitely integrate proper internationalization for this!

zzacharo (approved):
@mb-wali Thanks a lot for the PR!

zzacharo:
Merging for now until https://github.com/inveniosoftware/react-searchkit/issues/61 is addressed!

PR #174, opened 2021-02-04
pagination: add size option

zzacharo:


src/lib/components/Pagination/Pagination.js
```diff
@@ -13,10 +13,24 @@ import Overridable from 'react-overridable';
...
+  showNext: true,
+  size: 'large',
+};
+
 class Pagination extends Component {
   constructor(props) {
     super(props);
     this.updateQueryPage = props.updateQueryPage;
+    this.options = props.options
+      ? { ...defaultOptions, ...props.options }
```
zzacharo, line 32: This change was made because it's easier to override specific options e.g `options={size: 'small'}` without having to provide all the other keys.

ntarocco (approved):

PR #175, opened 2021-02-08
components: pass payload to resetQuery

zzacharo:
closes inveniosoftware/invenio-app-rdm#531

When the action `resetQuery` is triggered there is no way to choose how to reset the query state. This PR gives the resetQuery action a payload that can be passed from components that override e.g the EmptyResults component. By doing that you can control the default query after the reset action is triggered!

### DIFF
--- a/src/lib/api/contrib/invenio/InvenioSearchApi.js
+++ b/src/lib/api/contrib/invenio/InvenioSearchApi.js
@@ -11,7 +11,7 @@ import _get from 'lodash/get';
 import _hasIn from 'lodash/hasIn';
 import _isEmpty from 'lodash/isEmpty';
 import { updateQueryState } from '../../../state/selectors';
-import { INITIAL_QUERY_STATE } from '../../../storeConfig';
+import { INITIAL_QUERY_STATE_KEYS } from '../../../storeConfig';
 import { RequestCancelledError } from '../../errors';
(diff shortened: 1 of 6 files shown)


kprzerwa (approved):
LGTM but its better if someone else who worked more on the module also reviews :)

src/lib/api/contrib/invenio/InvenioSearchApi.js
```diff
@@ -98,7 +98,7 @@ export class InvenioSearchApi {
       const newQueryState = updateQueryState(
         stateQuery,
         response.extras,
-        INITIAL_QUERY_STATE
+        INITIAL_QUERY_STATE_KEYS
```
zzacharo, line 101: This was wrong and I fixed it....no clue how was working before...

src/lib/components/EmptyResults/index.js
```diff
@@ -11,7 +11,7 @@ import { resetQuery } from '../../state/actions';
 import EmptyResultsComponent from './EmptyResults';
 
 const mapDispatchToProps = (dispatch) => ({
-  resetQuery: () => dispatch(resetQuery()),
+  resetQuery: (resetState) => dispatch(resetQuery(resetState)),
```
ntarocco, line 14: The param `resetState` is not used?

ntarocco (approved):

PR #179, opened 2021-03-02
searchbar: responsive autofocus

jrcastro2:
* Adds new prop to add the autofocus
  depending on the width
* Needed by https://github.com/inveniosoftware/react-invenio-app-ils/pull/375


jrcastro2:
After discussion, this change doesn't belong to RSK

Issue #180, opened 2021-03-03
cosmetics: "Oups! Something went wrong while fetching results"


tiborsimko:
When something goes wrong during search, the error message currently says:

```
Oups! Something went wrong while fetching results
```

The problem: "Oups!" is not really English, even in slang. It reads a bit "franglais". 

Suggested alternatives: The text should rather say "Whoops!".  Or perhaps "Oops!".  Or perhaps "Ouch!", staying in the slang area still. Or perhaps "Oh no!", moving towards something more mainstream.

P.S. Yes, a totally cosmetic language bug report :wink:

PR #185, opened 2021-04-14
api-contrib: Add request serializer for invenio-records-resources

ParthS007:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/440
closes https://github.com/inveniosoftware/react-searchkit/issues/180

### DIFF
--- a/src/lib/api/contrib/invenio/InvenioSearchApi.js
+++ b/src/lib/api/contrib/invenio/InvenioSearchApi.js
@@ -15,6 +15,7 @@ import { INITIAL_QUERY_STATE_KEYS } from '../../../storeConfig';
 import { RequestCancelledError } from '../../errors';
 import { InvenioRequestSerializer } from './InvenioRequestSerializer';
 import { InvenioResponseSerializer } from './InvenioResponseSerializer';
+import { InvenioRecordsResourcesRequestSerializer } from './InvenioRecordsResourcesRequestSerializer';
 
 export class InvenioSearchApi {
   constructor(config) {
@@ -39,10 +40,14 @@ export class InvenioSearchApi {
   }
 
   initSerializers(config) {
+    this.requestSerializerMap = {
+      "InvenioRecordsResourcesRequestSerializer" : InvenioRecordsResourcesRequestSerializer,
+      "InvenioRequestSerializer" : InvenioRequestSerializer
+    }
     const requestSerializerCls = _get(
       config,
-      'invenio.requestSerializer',
-      InvenioRequestSerializer
+      this.requestSerializerMap[config['invenio.requestSerializer']],
+      InvenioRecordsResourcesRequestSerializer
     );
     const responseSerializerCls = _get(
       config,
--- a/src/lib/api/contrib/invenio/InvenioSearchApi.js
+++ b/src/lib/api/contrib/invenio/InvenioSearchApi.js
@@ -39,15 +39,21 @@ export class InvenioSearchApi {
     this.responseInterceptor = _get(config, 'interceptors.response', undefined);
   }
 
-  initSerializers(config) {
-    this.requestSerializerMap = {
-      "InvenioRecordsResourcesRequestSerializer" : InvenioRecordsResourcesRequestSerializer,
-      "InvenioRequestSerializer" : InvenioRequestSerializer
+  getSerializer(serializer) {
+    const requestSerializerMap = {
+      InvenioRecordsResourcesRequestSerializer: InvenioRecordsResourcesRequestSerializer,
+      InvenioRequestSerializer: InvenioRequestSerializer,
+    };
+    if (typeof serializer === 'string') {
+      return requestSerializerMap[serializer];
+    } else {
+      return serializer ? serializer : InvenioRequestSerializer;
     }
-    const requestSerializerCls = _get(
-      config,
-      this.requestSerializerMap[config['invenio.requestSerializer']],
-      InvenioRecordsResourcesRequestSerializer
+  }
+
+  initSerializers(config) {
+    const requestSerializerCls = this.getSerializer(
+      config.invenio?.requestSerializer
     );
     const responseSerializerCls = _get(
       config,
--- a/src/lib/api/contrib/invenio/InvenioRecordsResourcesRequestSerializer.js
+++ b/src/lib/api/contrib/invenio/InvenioRecordsResourcesRequestSerializer.js
@@ -1,0 +1,110 @@
+/*
+ * This file is part of React-SearchKit.
+ * Copyright (C) 2021 CERN.
+ *
+ * React-SearchKit is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import Qs from 'qs';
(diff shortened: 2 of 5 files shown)


src/lib/api/contrib/invenio-records-resources/InvenioRecordsResourcesRequestSerializer.js
```diff
@@ -0,0 +1,110 @@
...
+    this._checkFilter(filter)
+    const aggName = filter[0];
+    const fieldValue = filter[1];
+  	const hasChild = filter.length === 3;
+    if (hasChild) {
+      this._checkFilter(filter[2])
+    }
+    if (aggName in filterUrlParams) {
+      if (hasChild) {
+        filterUrlParams[aggName].push(fieldValue + "::" + filter[2][1])
```
ntarocco, line 42: I would make this `::` parametrizable if possible.

Generally speaking, I was trying to avoid this separators and let `Qs` to encode the string correctly. E.g. if you have lists, then it will be automatically encoded in some standard way, and decoded in the same way in the 
backend.
For example, you could have `['report', 'pdf']` instead of `report::pdf` and then see how Qs encodes it. In this way you don't need separators and you don't have the prob that maybe you have a value (very unlikely) `rep::ort`.

src/lib/api/contrib/invenio-records-resources/InvenioRecordsResourcesRequestSerializer.js
```diff
@@ -0,0 +1,110 @@
+/*
```
ntarocco, line 1: Can you please move this into the `invenio` contrib? It is just a new request serializer, not an entire new `contrib`.

src/lib/api/contrib/invenio-records-resources/InvenioRecordsResourcesRequestSerializer.js
ParthS007: I tried with pushing as list instead of separated by `::`, 

- the `filterUrlParams` looks like this:

```
resource_type: (3) […]
​​
0: Array [ "image", "image-photo" ]
​​
1: "software"
​​
2: Array [ "publication", "publication-thesis" ]
```

- the `q` looks like this:

```
q=&sort=newest&page=1&size=10&resource_type=image&resource_type=image-photo&resource_type=software&resource_type=publication&resource_type=publication-thesis
```

​​I am afraid we need to keep the separator since `image-photo`, `publication-thesis` are not resource type but `image::image-photo` and `publication::publication-thesisis` are resource_type with sub_type.

Another approach that I was thinking is passing the `splitchar` from the config but I then I have to access config here.

Please let me know which one sounds good.

src/lib/api/contrib/invenio/InvenioSearchApi.js
```diff
@@ -39,10 +40,14 @@ export class InvenioSearchApi {
...
+    this.requestSerializerMap = {
+      "InvenioRecordsResourcesRequestSerializer" : InvenioRecordsResourcesRequestSerializer,
+      "InvenioRequestSerializer" : InvenioRequestSerializer
+    }
     const requestSerializerCls = _get(
       config,
-      'invenio.requestSerializer',
-      InvenioRequestSerializer
+      this.requestSerializerMap[config['invenio.requestSerializer']],
+      InvenioRecordsResourcesRequestSerializer
```
ParthS007, line 50: Here, I am not sure how to decide which `Serializer` to put to check the type with.

src/lib/api/contrib/invenio-records-resources/InvenioRecordsResourcesRequestSerializer.js
ntarocco: Thanks for giving it a try, keep it like this then :)

src/lib/api/contrib/invenio/InvenioSearchApi.js
ntarocco: I am not sure what you mean. Here you should probably not change everything, it was already configurable.
What are you trying to achieve here?

It should have as default `InvenioRequestSerializer` otherwise it should be passed via config.

ParthS007: 1. The change at `line 49` is because of this error mentioned below, so I have to create a map to an actual function instead of a string.

When I am configuring the `data-invenio-search-config` in `search.html` for the new request adapter. I tried with these two configs:
- First: I get `Uncaught TypeError: requestSerializerCls is not a constructor` since here I am sending it as a string.
```
"searchApi": {
    "invenio": {
      "requestSerializer": "InvenioRecordsResourcesRequestSerializer"
    }
  }
```
- Second: I get `Uncaught SyntaxError: JSON.parse: unexpected character` since it is without quotes.
```
"searchApi": {
    "invenio": {
      "requestSerializer": InvenioRecordsResourcesRequestSerializer
    }
  }
```

2. For the `line 50`, if I am keeping `InvenioRequestSerializer`, this one is taken into consideration and the facets formats is like before. Please let me know if I am missing something here.

ntarocco: This is the right version indeed:

```
"searchApi": {
    "invenio": {
      "requestSerializer": InvenioRecordsResourcesRequestSerializer
    }
  }

```
but you cannot import it because you are in the `search.html`.
@zzacharo any idea how we can solve this? I would avoid the map, not flexible enough.

@ParthS007 I saw that in the `search.html` there is a spelling issue with Languages if you want to fix it:
```
      "title": "Laguanges"
```

ParthS007: okay, I will include this in my PR.

zzacharo: @ntarocco @ParthS007 There could be a solution to use `eval('ClassAsAString')` when we instantiate the `requestSerializer` in RSK (See https://stackoverflow.com/questions/49042459/how-to-instantiate-a-class-from-a-string-in-javascript). For that to work, you will need to import your custom class in the app-rdm bundle... That needs some experimentation though to see if the order of the bundle inclusion in the page is correct so RSK find the custom serializer class...

ntarocco: I wouldn't go for eval if possible, maybe dynamic import?

@ParthS007 maybe the best thing to do is:
1. in `invenio` contrib, create a new function that will (pseudocode):
```
function getSerializer (serializer) {
    if serializer is string:
        return mappings[serializer]
    else
         return serializer // it is a already a class
}
2. in the InvenioSearchApi.js, call `getSerializer(serializer)

So if you pass a string, will go through a mapping as you have already done, but if you pass already a class/obj, then it will simply instantiate it.
Does this make sense?

ParthS007: Yes, I think this sounds good and better for error handling as well. 

Pinging @jrcastro2, since he took over the issue now.

jrcastro2: Yes, I will go with this solution, thanks!

PR #187, opened 2021-04-15
EmptyResults: pass props down to rendered element

mvidalgarcia:
closes inveniosoftware/invenio-app-rdm#763

### DIFF
--- a/src/lib/components/EmptyResults/EmptyResults.js
+++ b/src/lib/components/EmptyResults/EmptyResults.js
@@ -28,12 +28,14 @@ class EmptyResults extends Component {
       queryString,
       extraContent,
       overridableId,
+      ...props
     } = this.props;
     return (
       <ShouldRender
         condition={!loading && _isEmpty(error) && totalResults === 0}
       >
         <Element
+          {...props}
           queryString={queryString}
           resetQuery={this.resetQuery}
           extraContent={extraContent}


src/lib/components/EmptyResults/EmptyResults.js
```diff
@@ -38,6 +39,7 @@ class EmptyResults extends Component {
           resetQuery={this.resetQuery}
           extraContent={extraContent}
           overridableId={overridableId}
+          {...props}
```
zzacharo, line 42: I would just say for an extra safety measure maybe to spread the `props` in the beginning of the `Element` component just to avoid passing a prop that overrides the defined ones... Is there any specific prop you wanted to pass? Maybe it's worth to pass only the properties defined for this component + specific redux ones?

mvidalgarcia: I want to pass a [custom prop](https://github.com/inveniosoftware/react-invenio-deposit/pull/261/files#diff-22d83991c42aee1d1e735a310b67cc2f4064a9157b6496fc7bbd375161af6258R164-R165), so I can use it in the [children element](https://github.com/inveniosoftware/react-invenio-deposit/pull/261/files#diff-e33cd2fd626c77a37e030fee4fe70b8836e089b4cf63e1deb0e90d6c8de7a258R25).

zzacharo: Ok then I would say just upack the props as first thing to avoid just potential unwanted overrides :)

ntarocco (approved):

Issue #183, opened 2021-03-26
Reset button for SearchBar


ksuess:
## Describe the solution you'd like
An optional reset button to empty the search field would be nice.

## Describe alternatives you've considered
Maybe you have a hint for overriding this component? Im stuck with addressing the store; updateQueryState?

jrcastro2:
Thanks a lot for the suggestion, one possible solution would be adding a `X` icon at the end. 
![image](https://user-images.githubusercontent.com/15194802/115358787-4ea18b00-a1be-11eb-97b8-cdef3fd436ad.png)
What do you think about this solution? 

As for overriding the `Searchbar` you would need to override the `Input` component (check code [here](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SearchBar/SearchBar.js#L128-L150)) using `react-overridable` (documentation with examples available [here](https://github.com/indico/react-overridable)).

ksuess:
Thanks for your response. I do like your suggested solution / mockup.
I wish I could have already placed a PR for this. But I need some more experience with react-overridable. Despite I already achieved to override some other components and got a really nice looking faceted search for https://github.com/plone/volto Thanks a lot for this great searchkit package! Hope to give something back…


May I also suggest to make the x-Icon overridable. I think this is important for theming.

![image](https://user-images.githubusercontent.com/1144737/115757941-e3fb7580-a39e-11eb-815a-d93cc947a9c3.png)

ntarocco:
Thank  you! It looks like we can close this issue now :)

PR #190, opened 2021-04-25
SearchBar: Reset button

ksuess:
Implementation of https://github.com/inveniosoftware/react-searchkit/issues/183


ksuess:
Problem is that uiProps are not applied anymore, as <Input> is passed children.

PR #202, opened 2021-09-20
fixed bug of pagination disappearing after a custom query state update

NRodriguezcuellar:
the pagination menu was disappearing after using the custom query build helper of ILS. After investigation i learned that this was because it defaulted back to the initial state if no page number or size was passed while updating the state. It now takes the current values if those 2 properties are not passed with the payload.

closes https://github.com/inveniosoftware/react-invenio-app-ils/issues/344

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+Version 1.0.0-alpha.17 (released 2021-07-27)
+
+* Fix deselection of nested facet.
+  Fixes https://github.com/inveniosoftware/invenio-app-rdm/issues/1014
+
 Version 1.0.0-alpha.16 (released 2021-07-21)
 
 * Update axios dep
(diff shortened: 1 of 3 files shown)


src/lib/state/actions/query.js
```diff
@@ -52,10 +52,12 @@ export const onAppInitialized = (searchOnInit) => {
...
 
 export const updateQueryState = (queryState) => {
-  return (dispatch) => {
+  return (dispatch, getState) => {
+    const {query: {size, page}} = getState()
+
     dispatch({
       type: SET_QUERY_STATE,
-      payload: queryState,
+      payload: {size, page, ...queryState},
```
ntarocco, line 60: Question: would it be safer or more correct to this the following instead?
```suggestion
      const { query: currentQueryState } = getState()
      payload: {...currentQueryState, ...queryState},
```

NRodriguezcuellar: i think it in terms of safeness there is no big difference, because it both cases the new queryState object will override either the whole currentQueryState or just the page and size properties, if it has overlapping properties🤔

PR #208, opened 2022-01-25
global: upgrade npm deps

ntarocco:
Could this upgrade have any conflict with InvenioRDM?


zzacharo (approved):
LGTM! I think is not breaking something for RDM

PR #212, opened 2022-02-01
change the test adapter to support react 17

### LINKED ISSUE #211: Support React 17
To allow using searchkit in projects on React 17, we would need to upgrade `react` and `react-dom` dev and peer dependencies to ^17 and ensure test still run with success. 

Anything else we should check?

chelakis:
Closes #211

Changed the adapter to work with react 17, changed react, react-dom versions and run the tests again with success


ntarocco:
Thank you for this PR. We are discussing how to release it, because we cannot force `react` `v17` at the moment in many other projects.

PR #214, opened 2022-02-04
Toggle: add namespace to overridable id

jrcastro2:


src/lib/components/Toggle/Toggle.js
```diff
@@ -65,6 +30,46 @@ ToggleComponent.defaultProps = {
   overridableId: '',
 };
 
+const Element = ({ overridableId, ...props }) => {
```
zzacharo, line 33: I think this is redundant...I guess you did it to use the `useContext` hook. If so, what about just defining the `ToggleComponent` class to a function?

zzacharo (commented):
Thanks for that! It seems we missed namespacing that component!

src/lib/components/Toggle/Toggle.js
jrcastro2: Fair point. Not sure what I was thinking to keep the class component :sweat_smile:. Thanks!

zzacharo (approved):

PR #215, opened 2022-02-07
A11y: Sort dropdown ARIA label

lhenze:
Adding possibility for accessible name to the search results sort dropdown

### DIFF
--- a/src/lib/components/Sort/Sort.js
+++ b/src/lib/components/Sort/Sort.js
@@ -1,6 +1,7 @@
 /*
  * This file is part of React-SearchKit.
- * Copyright (C) 2018-200 CERN.
+ * Copyright (C) 2018-2022 CERN.
+ * Copyright (C) 2022 NYU.
  *
  * React-SearchKit is free software; you can redistribute it and/or modify it
  * under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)


src/lib/components/Sort/Sort.js
```diff
@@ -83,6 +86,7 @@ Sort.propTypes = {
   label: PropTypes.func,
   overridableId: PropTypes.string,
   sortOrderDisabled: PropTypes.bool,
+  ariaLabel: PropTypes.string
```
ntarocco, line 89: can you please add the `default` value just below, given that it is not a required `prop`?

ntarocco (commented):
Thanks for the contribution!
Are you planning to add the `ARIA` tags to more components?

src/lib/components/Sort/Sort.js
lhenze: Empty string is not great for A11y.  I put in the default value of "Sort" even though it's not translated.  Will also make a PR for invenio-search-ui that includes translation option.

NRodriguezcuellar (approved):

kpsherva (commented):
It looks good to me!
Could you squash the commits to one and edit the commit message in order to follow the invenio contribution guidelines?

lhenze:
Commits have been squashed and message amended.

PR #227, opened 2022-04-21
toggle: update exported overridable id

jennur:
Closes https://github.com/inveniosoftware/react-invenio-deposit/issues/469

### DIFF
--- a/src/lib/components/Toggle/Toggle.js
+++ b/src/lib/components/Toggle/Toggle.js
@@ -66,4 +66,4 @@ ToggleComponent.defaultProps = {
   overridableId: "",
 };
 
-export default Overridable.component("SearchFilters.Toggle", ToggleComponent);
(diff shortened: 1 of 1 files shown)


src/lib/components/Toggle/Toggle.js
```diff
@@ -66,4 +66,4 @@ ToggleComponent.defaultProps = {
   overridableId: "",
 };
 
-export default Overridable.component("SearchFilters.Toggle", ToggleComponent);
+export default Overridable.component("SearchFilters.Toggle.element", ToggleComponent);
```
kpsherva, line 69: wy is this change needed? it is not consistent with the rest of the components ids and potentially a breaking change

jennur:
Should be updated in https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/components/License/LicenseModal.js#L30 instead.

PR #249, opened 2023-03-30
pagination: add maxTotalResults prop

jennur:
Closes https://github.com/zenodo/zenodo-rdm/issues/217

Adds `maxTotalResults` as part of the `options` prop in `<Pagination/>`

### DIFF
--- a/src/lib/components/Pagination/Pagination.js
+++ b/src/lib/components/Pagination/Pagination.js
@@ -53,6 +53,7 @@ class Pagination extends Component {
           onPageChange={this.onPageChange}
           options={this.options}
           overridableId={overridableId}
+          maxTotalResults={this.options.maxTotalResults}
         />
       </ShouldRender>
     );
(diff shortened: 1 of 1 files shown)


src/lib/components/Pagination/Pagination.js
```diff
@@ -69,6 +70,7 @@ Pagination.propTypes = {
     showPrev: PropTypes.bool,
     showNext: PropTypes.bool,
     size: PropTypes.oneOf(["mini", "tiny", "small", "large", "huge", "massive"]),
+    maxTotalResults: PropTypes.number,
```
kpsherva, line 73: do we still keep the default here: https://github.com/inveniosoftware/react-searchkit/pull/249/files#diff-5e6e91a6aa8b7da0df03350d588f4d4af38b55cbf4de505be53fb628e98304cfR155 ?

ppanero: I think is safer to have a default, then if the backend wants to modify it can do. WDYT?

ppanero (approved):

PR #250, opened 2023-07-12
resultsList: add overridable shouldRender

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2180

### DIFF
--- a/src/lib/components/ResultsList/ResultsList.js
+++ b/src/lib/components/ResultsList/ResultsList.js
@@ -14,10 +14,22 @@ import { AppContext } from "../ReactSearchKit";
 import { ShouldRender } from "../ShouldRender";
 
 function ResultsList({ loading, totalResults, results, overridableId }) {
+  const { buildUID } = useContext(AppContext);
+
   return (
-    <ShouldRender condition={!loading && totalResults > 0}>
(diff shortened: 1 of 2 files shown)


src/lib/components/ResultsList/ResultsList.js
```diff
@@ -14,10 +14,22 @@ import { AppContext } from "../ReactSearchKit";
...
 
 function ResultsList({ loading, totalResults, results, overridableId }) {
+  const { buildUID } = useContext(AppContext);
+
   return (
-    <ShouldRender condition={!loading && totalResults > 0}>
-      <Element results={results} overridableId={overridableId} />
-    </ShouldRender>
+    <Overridable
+      id={buildUID("ResultsList.ShouldRender.container", overridableId)}
```
jrcastro2, line 21: Why did I add this new id? 
Because the `ResultsList.container` is already being used (you can check it down in this file). I would be more in favour of namig this `ResultsList.container` and removing the `Element` cmp defined down, but by doing so this would be a breaking change and we would need to update all the places we are using this, I am glad to do this but I would like some confirmation before, ping @zzacharo

zzacharo:
discussed IRL that this is not needed anymore

PR #257, opened 2023-11-01
InvenioSerializer: Remove parent category in nested filters

psaiz:

### Description

The nested facets are not working properly. A request like `f=type%3ADataset&f=type%3ADocumentation%2Bsubtype%3AAbout` gets translated into `type=Dataset&type=Documentation&subtype=About`, instead of `type=Dataset&subtype=About`.

Note that the comments in the code (line 45-60) were correct. The implementation was different :-)


psaiz:
Changed a bit to make it more similar to the RDM. Instance of putting only the child category, it concatenates the value of the parent and the child with '::'

ntarocco (approved):

PR #260, opened 2023-12-11
results grid: pass correct value to results prop

jennur:
Closes https://github.com/zenodo/rdm-project/issues/567

- I believe this was a bug, as it is passed in the same way for `ResultsList`, and also makes it easier to override this way.

### DIFF
--- a/src/lib/components/ResultsGrid/ResultsGrid.js
+++ b/src/lib/components/ResultsGrid/ResultsGrid.js
@@ -71,7 +71,7 @@ const Element = ({ overridableId, results, resultsPerRow }) => {
   return (
     <Overridable
       id={buildUID("ResultsGrid.container", overridableId)}
-      results={results}
+      results={_results}
       resultsPerRow={resultsPerRow}
     >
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

jrcastro2 (approved):

src/lib/components/ResultsGrid/ResultsGrid.js
```diff
@@ -71,7 +71,7 @@ const Element = ({ overridableId, results, resultsPerRow }) => {
   return (
     <Overridable
       id={buildUID("ResultsGrid.container", overridableId)}
-      results={results}
+      results={_results}
```
ntarocco, line 74: @kpsherva @jrcastro2 given that you have reviewed this one.
IMO, I don't think this was 100% correct: the idea here is to pass the "raw" values, that would allow you to use `results` to build your own layout when overriding.
Passing `_results` means that when overriding, you get already a `GridItem` React cmp, instead of the list of `results` object, and you cannot easily change that cmp.

Can you please fix this?

Ping @zzacharo for any other opinion.

Issue #258, opened 2023-11-02
State selection when different options start with the same name


psaiz:
Package version (if known): 2.0.0
## Describe the bug

I have a checkbox with multiple options. If the name of one of the options is a subset of a different one, then, the selection does not work properly. 
## Steps to Reproduce
Look at the screenshot, where there are several options: 'Configuration', 'Configuration HLT'

1. Click on 'Configuration HLT', and things work as expected (the checkbox gets selected)
2. Click on 'Configuration', and the checkbox of 'Configuration HLT' gets unselected.

## Expected behavior

After the second click, both 'Configuration' and 'Configuration HLT' should be selected. 

This is probably related to [this line](https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/state/selectors/query.js#L98), and the comparison with `startswith`.

## Additional context
![Screenshot 2023-11-02 at 16 16 11](https://github.com/inveniosoftware/react-searchkit/assets/12909623/8e757d84-87dc-4f1a-8572-00b25f0a9cbc)

PR #261, opened 2024-01-15
selector: check the corner case when one value starts with the same s…

psaiz:
…tring as another value

### Description
Addresses https://github.com/inveniosoftware/react-searchkit/issues/258

### DIFF
--- a/src/lib/state/selectors/query.js
+++ b/src/lib/state/selectors/query.js
@@ -18,7 +18,16 @@ import _pick from "lodash/pick";
  * @param {string} second a string
  */
 function startsWith(first, second) {
-  return first.indexOf(second) === 0;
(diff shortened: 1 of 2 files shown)


src/lib/state/selectors/query.js
```diff
@@ -18,7 +18,16 @@ import _pick from "lodash/pick";
  * @param {string} second a string
  */
 function startsWith(first, second) {
-  return first.indexOf(second) === 0;
+  /* This is a bit tricky, since the verification should distinguish between the label and the children
+  In other words:
+  * 'q=type:hello&subtype:there' (or 'q=type:hello') starts with 'q=type:hello'
```
zzacharo, line 23: Nice test cases! Can you include them in the test?

ntarocco (approved):

zzacharo (approved):

Issue #259, opened 2023-11-24
Nested filters with the same value


psaiz:
Package version (if known): 2.2.4

## Describe the bug

I have a category filter that looks like the attachment. There is a nested field, with values like 'Environment', 'Software', 'News', and then each one might have different children values. Note that the children value of 'Validation' appears in two parents ('Environment' and 'Software'). Clicking on any of the 'Validation' children selects the first one  ('Environment'). There is no way of selecting 'Software::Validation'. 

## Steps to Reproduce
1. Have a nested field, and the same value on two children. 
2. Check if the second one can be selected
 
## Expected behavior

The user should be able to select the first or the second one

## Screenshots (if applicable)

![Screenshot 2023-11-24 at 16 19 17](https://github.com/inveniosoftware/react-searchkit/assets/12909623/e4da188c-bce5-40ec-8f71-7e12adc356a4)

PR #262, opened 2024-01-15
selector: Allow the same value in multiple children aggregations

psaiz:

### Description
Closes https://github.com/inveniosoftware/react-searchkit/issues/259

### DIFF
--- a/src/lib/components/BucketAggregation/BucketAggregationValues.js
+++ b/src/lib/components/BucketAggregation/BucketAggregationValues.js
@@ -134,12 +134,14 @@ const ValueElement = ({
   overridableId,
   keyField,
 }) => {
-  const { buildUID } = useContext(AppContext);
+  const { buildUID, nextComponentIndex } = useContext(AppContext);
   const label = bucket.label
     ? bucket.label
     : `${keyField} (${bucket.doc_count.toLocaleString("en-US")})`;
   const childAggCmps = getChildAggCmps(bucket);
-
+  /* Note that the checkbox does not have any `id`. It used to have one, based on the value, and the issue
+     was that, if the same value appeared more than once, react didn't like having two items with the same id,
+     and clicking on the second would select the first one */
   return (
     <Overridable
       id={buildUID("BucketAggregationValues.element", overridableId)}
@@ -152,7 +154,7 @@ const ValueElement = ({
       <List.Item key={bucket.key}>
         <Checkbox
           label={label}
-          id={`${bucket.key}-agg-value-checkbox`}
+          id={nextComponentIndex()}
           value={bucket.key}
           onClick={() => onFilterClicked(bucket.key)}
           checked={isSelected}
--- a/src/lib/components/ReactSearchKit/ReactSearchKit.js
+++ b/src/lib/components/ReactSearchKit/ReactSearchKit.js
@@ -35,8 +35,8 @@ export class ReactSearchKit extends Component {
     this.store = createStoreWithConfig(appConfig);
     this.appName = props.appName;
     this.eventListenerEnabled = props.eventListenerEnabled;
+    this.componentIndex = 0;
   }
-
   render() {
     const { appName, children, eventListenerEnabled, overridableId, searchOnInit } =
       this.props;
@@ -44,6 +44,9 @@ export class ReactSearchKit extends Component {
     const context = {
       appName,
       buildUID: (element, overrideId) => buildUID(element, overrideId, appName),
+      /* This is an internal function that will be used to create id for the checkboxes and labels, ensuring that
+          they are unique */
+      nextComponentIndex: () => `${this.appName}_${this.componentIndex++}`,
     };
 
     const _overridableId = buildUID("ReactSearchKit.children", overridableId, appName);
--- a/src/lib/state/selectors/query.test.js
+++ b/src/lib/state/selectors/query.test.js
@@ -154,6 +154,7 @@ describe("queries with second level filters.", () => {
       ["type", "Publication"],
     ]);
   });
+
   test("query with terms that start with the same text and with children", () => {
     const state = [
       ["file_type", "pdf", ["sub_type", "image"]],
@@ -164,6 +165,41 @@ describe("queries with second level filters.", () => {
     const newState = updateQueryFilters(query, state);
 
     expect(newState).toEqual([["file_type", "pdf2"]]);
+  });
+
+  test("query with `subtype: Validation` should add it to the correct type", () => {
+    const state = [
+      ["file_type", "pdf"],
+      ["type", "Documentation", ["subtype", "Validation"]],
+    ];
+
+    const query = ["type", "Publication", ["subtype", "Validation"]];
+
+    const newState = updateQueryFilters(query, state);
+
+    expect(newState).toEqual([
+      ["file_type", "pdf"],
+      ["type", "Documentation", ["subtype", "Validation"]],
+      ["type", "Publication", ["subtype", "Validation"]],
+    ]);
+  });
+
+  test("query with `subtype: Validation` should remove it from the correct type", () => {
+    const state = [
+      ["file_type", "pdf"],
+      ["type", "Documentation", ["subtype", "Validation"]],
+      ["type", "Publication", ["subtype", "Validation"]],
+    ];
+
+    const query = ["type", "Publication", ["subtype", "Validation"]];
+
+    const newState = updateQueryFilters(query, state);
+
+    expect(newState).toEqual([
+      ["file_type", "pdf"],
+      ["type", "Documentation", ["subtype", "Validation"]],
+      ["type", "Publication"],
+    ]);
   });
 });
 


src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -97,11 +100,13 @@ BucketAggregationValues.propTypes = {
   }),
   onFilterClicked: PropTypes.func.isRequired,
   overridableId: PropTypes.string,
+  parentValue: PropTypes.string,
 };
 
 BucketAggregationValues.defaultProps = {
   childAgg: null,
   overridableId: "",
+  parentValue: "",
```
ntarocco, line 109: Isn't it this required now? If yes, default props should appear only for optional arguments:
```suggestion

```

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -97,11 +100,13 @@ BucketAggregationValues.propTypes = {
   }),
   onFilterClicked: PropTypes.func.isRequired,
   overridableId: PropTypes.string,
+  parentValue: PropTypes.string,
```
ntarocco, line 103:
```suggestion
  parentValue: PropTypes.string.isRequired,
```

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -170,10 +176,12 @@ ValueElement.propTypes = {
   getChildAggCmps: PropTypes.func.isRequired,
   keyField: PropTypes.string.isRequired,
   overridableId: PropTypes.string,
+  parentValue: PropTypes.string,
```
ntarocco, line 179:
```suggestion
  parentValue: PropTypes.string.isRequired,
```

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -170,10 +176,12 @@ ValueElement.propTypes = {
...
 };
 
 ValueElement.defaultProps = {
   overridableId: "",
+  parentValue: "",
```
ntarocco, line 184:
```suggestion

```

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -53,6 +54,7 @@ class BucketAggregationValues extends Component {
           aggName={this.childAgg.aggName}
           childAgg={this.childAgg.childAgg}
           onFilterClicked={onFilterClicked}
+          parentValue={bucket.key}
```
ntarocco, line 57: Very minor: I think that we tend to keep the `key` naming around.
Shall we call this `parentKey` or `parentBucketKey` instead?

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -19,6 +19,7 @@ class BucketAggregationValues extends Component {
     this.aggName = props.aggName;
     this.childAgg = props.childAgg;
     this.onFilterClicked = props.onFilterClicked;
+    this.parentValue = props.parentValue;
```
ntarocco, line 22: I have the feeling that you might have to change the cmp `BucketAggregation`, to pass the initial value for this prop.

src/lib/state/selectors/query.test.js
```diff
@@ -142,6 +142,23 @@ describe("queries with second level filters.", () => {
       ["type", "Publication"],
     ]);
   });
+
+  test("query with `subtype: Validation` should remove it from the correct type", () => {
```
zzacharo, line 146: I think you test that you add `Validation` to the correct type here no? Either way, also testing the removal is good I think.

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -152,7 +158,7 @@ const ValueElement = ({
       <List.Item key={bucket.key}>
         <Checkbox
           label={label}
-          id={`${bucket.key}-agg-value-checkbox`}
+          id={`${parentValue}-${bucket.key}-agg-value-checkbox`}
```
zzacharo, line 161: question: I assume that this is the fix differentiating the components, can you point me to the code that utilizes this?

ntarocco: What I believe is happening (to be confirmed) is that `React` is adding an event listener (click handler) on the `id` of the HTML tag. If you look how it is rendered, there is no inline `onClick` or similar:
`<input class="hidden" id="Validation-agg-value-checkbox" readonly="" tabindex="0" type="checkbox" value="Validation">`

This means that is another checkbox has the same `id`, ids become duplicated and the first is picked.
Having unique ids looks like fixes the issue.

zzacharo: ah right, I thought it was somewhat connected with our internal implementation of filtering. It makes sense, thanks for the explanation! Maybe a neat comment @psaiz: 
```suggestion
          id={parentValue ? `${parentValue}-${bucket.key}-agg-value-checkbox` : `${bucket.key}-agg-value-checkbox`}
```

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: We are going for a different solution, where we don't specify any more the id of the element. There is no need to pass the parent anymore

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: We are going for a different solution, where we don't specify any more the id of the element. There is no need to pass the parent anymore

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: We are going for a different solution, where we don't specify any more the id of the element. There is no need to pass the parent anymore

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: We are going for a different solution, where we don't specify any more the id of the element. There is no need to pass the parent anymore

src/lib/state/selectors/query.test.js
psaiz: Good point. Added.

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: Ok, comment added

src/lib/components/BucketAggregation/BucketAggregationValues.js
psaiz: We are going for a different solution, where we don't specify any more the id of the element. There is no need to pass the parent anymore

src/lib/components/BucketAggregation/BucketAggregationValues.js
```diff
@@ -152,7 +154,6 @@ const ValueElement = ({
       <List.Item key={bucket.key}>
```
zzacharo, line 154: Given that the id was the same as the `key`, then the key value is also wrong in that case as there is duplicate value.

psaiz: This one does not seem to cause an issue. Note that they are on different branches.

src/lib/components/ReactSearchKit/ReactSearchKit.js
```diff
@@ -35,15 +35,21 @@ export class ReactSearchKit extends Component {
...
   }
-
   render() {
     const { appName, children, eventListenerEnabled, overridableId, searchOnInit } =
       this.props;
 
     const context = {
       appName,
       buildUID: (element, overrideId) => buildUID(element, overrideId, appName),
+      nextComponentIndex: () => this.nextComponentIndex(this),
```
psaiz, line 52: Please, check this! I started with something like :
```
  nextComponentIndex: this.nextComponentIndex,
```
and that failed because then `this` was empty inside the `nextComponentIndex` function. If there is a better way of making sure that `this` is available, please, let me know

ntarocco: There are two things happing here:
- the first is the always confusing usage of `this` in js
- the second is that the way you are testing (by changing the code in the compiled version of this JS) **might** differ from what the JS/babel compiler will do. The usage of `this` might be compiled differently. It might be the same, but it might be not.

To be sure, you can run the demo apps inside React-SearchKit, following the README, so you have a local running app to test.

Given said so, I think that your previous version should work:
```javascript
getNextUniqueId() {
    return `${this.appName}-${this.nextUniqueId++}`;
}
```

psaiz: Fixed, thanks

ntarocco (approved):

PR #283, opened 2025-05-02
update axios dependency and update tests

Samk13:
related to https://github.com/inveniosoftware/invenio-search-ui/pull/215

- Update axios deps 1.8.2

- Add `TextEncoder`, `TextDecoder`, and `ReadableStream` as globals to address jsdom/jsdom#2524. Update `axios` and `axios-mock-adapter` versions in `package.json`, pin `cheerio` due to an issue with enzyme, and modify Jest test command to transform `node_modules/axios`. Fix tests to accommodate these changes.

# Question:
should we merge this, or bump the upper pin here: https://github.com/inveniosoftware/invenio-rdm-records/blob/maint-10.x/invenio_rdm_records/webpack.py#L40C38-L40C44


package.json
```diff
@@ -17,14 +17,14 @@
     "postlink-dist": "cd dist && rm -rf node_modules",
     "unlink-dist": "cd dist && npm unlink && rm package*",
     "watch": "NODE_ENV=development rollup --watch -c",
-    "test": "react-scripts test",
+    "test": "react-scripts test --transformIgnorePatterns /\"node_modules/(?!axios)/\"",
     "eject": "react-scripts eject",
-    "lint": "eslint src/ --ext .js --max-warnings=0",
+    "lint": "eslint src/ --ext .js --max-warnings=5",
```
Samk13, line 22: we have 3 warnings here: `warning  Assignments to the 'onResultsRendered' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect`

.github/workflows/js-tests.yml
```diff
@@ -19,18 +26,18 @@ on:
 
 jobs:
   Tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [22.x, 18.x, 16.x]
```
utnapischtim, line 33: end of life of 16.x and 18.x is in the past [see](https://endoflife.date/nodejs) i think we can remove them!

.github/workflows/js-tests.yml
```diff
@@ -1,12 +1,19 @@
...
+# Copyright (C) 2024 CERN.
+# Copyright (C) 2025 KTH Royal Institute of Technology.
+
 name: CI
 
 on:
   push:
     branches:
       - master
+      - maint-*
```
utnapischtim, line 12: i am not sure, but in the other packages we did a "- maint-**", so with double quotes. i am not sure if this works

.github/workflows/js-tests.yml
```diff
@@ -19,18 +26,18 @@ on:
 
 jobs:
   Tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
utnapischtim, line 29: should we change to ubuntu-latest?

utnapischtim (approved):

.github/workflows/js-tests.yml
Samk13: I only kept them to match master: https://github.com/inveniosoftware/react-searchkit/blob/master/.github/workflows/js-tests.yml#L26
We need to also update react-forms in that case: https://github.com/inveniosoftware/react-invenio-forms/blob/master/.github/workflows/js-tests.yml
and reflect this change on master as well.

PR #296, opened 2026-04-02
Enable translations

SarahW91:
### Description

I think it would be beneficial to enable translations for this module. There are some rather prominent components, e.g. `EmptyResults` that can't be translated otherwise. 

Adding the necessary scripts and dependencies adds this ability, e.g.:

image

What do you think? If you are generally okay with this, I'll start marking all strings in the module as translatable.

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commit-message).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).


src/lib/translations/scripts/compileCatalog.js
```diff
@@ -0,0 +1,38 @@
+// This file is part of React-Searchkit
```
Samk13: Could we align the scripts with the implementation we already use in invenio-app-rdm? https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/scripts/compileCatalog.js

src/lib/translations/messages/index.js
```diff
@@ -0,0 +1,69 @@
+import TRANSLATE_AF from "./af/translations.json";
```
Samk13: Same here, check the PR: https://github.com/inveniosoftware/invenio-app-rdm/commit/13c14901dbcb4b14956a378802befac79de3a7f3#diff-8b01847cace59a05dffd86a564f110ec278d9f3913752994b04ed6c42ec72add

Samk13 (changes requested):
Thanks for this @SarahW91! this is a nice addition, enabling translations definitely makes sense.

Before merging, I'd prefer we align this a bit more with what we already have in [invenio-app-rdm](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/package.json):

- use the more recent [compileCatalog.js](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/scripts/compileCatalog.js) pattern (async/await, proper error handling, normalized output, etc..),
- generate the translations index (like [_generatedTranslations.js](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/messages/_generatedTranslations.js)) instead of maintaining messages/index.js manually.

This will keep things consistent across packages and avoid extra maintenance later on.

src/lib/translations/messages/index.js
SarahW91: What I don't quite get is where the *.po files and language configs are coming from. Are they pulled from Transifex? because there seems to be no step to automatically generate them otherwise

Samk13: Yes, without Transifex yet, the new language has to be bootstrapped locally first.

In the current Invenio setup, [init_catalog](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/scripts/initCatalog.js)  only adds the language to package.json; it does not create the .po file automatically.
Then [extract_messages](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/assets/semantic-ui/translations/invenio_app_rdm/package.json#L13) updates the extracted JSON and generates the English source files (translations.pot / messages/en/messages.po).
For any additional language, you still need to create messages/<lang>/messages.po locally first, and only then compile_catalog can convert it to translations.json.

So the local dev flow is basically:
```sh
npm run init_catalog lang <lang>
npm run extract_messages # create/bootstrap messages/<lang>/messages.po
npm run compile_catalog lang <lang>
```
Once Transifex is set up, that manual step (dev work) gets replaced by pulling the .po files from Transifex.

It makes me wonder where all these languages listed in this PR are coming from. Were they generated automatically, or are they copied manually from another package?

SarahW91: I copied those over from https://github.com/inveniosoftware/react-invenio-deposit

SarahW91: I updated the language list to be the same as on Transifex, created all *.po files and "compiled" the translations to generate the translation.json files. 

What would still be missing now to merge this PR @Samk13 ?

.tx/config
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (C) 2019 CERN.
+# Copyright (C) 2019 Northwestern University.
+# Copyright (C) 2023 Graz University of Technology.
+#
+# React-Searchkit is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+# Translate JavaScript strings
+# 1) Navigate to the directory:
+#    cd react-searchkit/src/lib/translations/messages
```
Samk13, line 12:
```suggestion
#    cd src/lib/translations/messages
```

src/lib/translations/messages/index.js
```diff
@@ -0,0 +1,24 @@
...
+import TRANSLATE_en from "./en/translations.json";
+import TRANSLATE_es from "./es/translations.json";
+export const translations = {
+  de: { translation: TRANSLATE_de },
+  en: { translation: TRANSLATE_en },
+  es: { translation: TRANSLATE_es },
+};
+*/
+
+import { translations } from "./_generatedTranslations";
```
Samk13, line 22: Without this file `_generatedTranslations` that include the languages you have in package.json, the installation will not proceed correctly, so you need to push it as well.

package.json
```diff
@@ -8,6 +8,41 @@
...
+  "config": {
+    "languages": [
+      "ar",
+      "bg",
+      "ca",
+      "cs",
+      "da",
+      "de",
+      "el",
+      "en",
```
Samk13, line 20: After we test that it's working, we can remove all these, and keep `en` for now, when we set up the Transifex repo, it will update automatically

src/lib/translations/messages/en/messages.po
```diff
@@ -0,0 +1,55 @@
+msgid ""
```
Samk13, line 1: I think it's safe to delete all other languages other than `en` at this stage. Once we set up the Transifex, then we can do another pass.

package.json
```diff
@@ -112,5 +160,10 @@
   },
   "engines": {
     "node": ">=14.0.0"
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "@translations/i18next": "<rootDir>/src/lib/i18next"
+    }
   }
```
Samk13, line 168: Should we remove this block? 
It looks like it was added only to mirror the new Rollup alias for `@translations/i18next`, so Jest can resolve that import if tests cover those components. 
I don’t think it’s exercised by the current test suite, so if we keep the alias it’s mostly defensive; if we drop the alias and use relative imports instead, this config can go away too.
WDYT?

Samk13 (commented):
Thanks for this @SarahW91. I left few comments on smaller cleanup points, but from my testing, the PR looks good and seems ready to merge.


Image

We will have another pass once we publish it on Transifex if something needs to be adjusted, as we can't push directly while developing.

And there are many places where the strings are not wrapped for translations, but you can choose if we add it here or in a follow-up PR e.g:

- https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/Sort/Sort.js#L159

- https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SortBy/SortBy.js#L126
- https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SortOrder/SortOrder.js#L128
- https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/ResultsPerPage/ResultsPerPage.js#L81
- https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/RangeFacet/RangeCustomFilter.js#L369-L372

SarahW91:
> And there are many places where the strings are not wrapped for translations, but you can choose if we add it here or in a follow-up PR e.g:
> 
>     * https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/Sort/Sort.js#L159
> 
>     * https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SortBy/SortBy.js#L126
> 
>     * https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/SortOrder/SortOrder.js#L128
> 
>     * https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/ResultsPerPage/ResultsPerPage.js#L81
> 
>     * https://github.com/inveniosoftware/react-searchkit/blob/master/src/lib/components/RangeFacet/RangeCustomFilter.js#L369-L372

I'm not sure if AriaLabels need translating. Where do they show up for users? Sorry if that's obvious, but I'm not usually a frontend dev ;)

Samk13:
> 
> I'm not sure if AriaLabels need translating. Where do they show up for users? Sorry if that's obvious, but I'm not usually a frontend dev ;)

hmmm, I'd lean toward translating them. ARIA labels don't show up visually in the UI, but they're used by screen readers, so they're still part of the user experience for people relying on assistive technologies.

If those users don't speak English, leaving ARIA labels untranslated can make navigation difficult or confusing. From an accessibility perspective, its usually best to include them in translations.

That said, its not always enforced strictly all over our code base, so as I mentioned, we could also handle it in a follow up if thats easier.

package.json
SarahW91: I don't know enough about this setup to feel confident in removing it. If someone with more experience thinks it's safe, I can do it, but for now I'd say we keep it.

package.json
SarahW91: But there's no harm in keeping them for now? I'd rather remove them in a separate PR after we see everything working together with Transifex

SarahW91: But seems to work:

image

_generated_translations.js also looks fine to me, I'll push it soon

Samk13:
Thanks for this, @SarahW91! I've squashed and cleaned up this one in 305, pushed everything to Transifex, so we can close this PR now.
see: https://app.transifex.com/inveniosoftware/invenio/react-searchkit-messages-ui/

PR #297, opened 2026-04-10
feat(RangeFacet): improve custom date filter UX

slint:
> AI use disclosure: (Obviously) I implemented the majority of this with Claude Code Opus 4.6. Had to guide a bit for backwards compat and keeping in line with the rest of the codebase style. Otherwise, the main focus was on UX, so I thoroughly tested expected behavior and display on my local instance.

* Replace multi-input date fields with single masked inputs that
  auto-format as YYYY, YYYY-MM, or YYYY-MM-DD
* Remove DateRangeInputs component (replaced by masked inputs)
* Add helptext hint below custom range inputs
* Add search icon button and Enter key to apply custom range
* Support open-ended ranges (empty lower or upper bound)
* Add input validation with error feedback on the inputs
* Change default filter checkboxes to radio buttons
* Add focus-to-activate behavior on custom range inputs
* Unify default and custom filters into a single radio group
* Add valueType prop ("date" or "int") for generic range support
* Export VALUE_TYPES from utils
* Remove unused customDatesLabel and datePlaceholders props

---

### Before (🔉 sound on!)

https://github.com/user-attachments/assets/74ea5f7e-5d4d-4d94-8d6c-d5e19c3240e2

### After (🔉 sound on!)

https://github.com/user-attachments/assets/b78445d0-d978-4b82-a7de-945839a67c3b


src/lib/components/RangeFacet/RangeCustomFilter.js
```diff
@@ -70,213 +85,246 @@ class RangeCustomFilter extends React.Component {
...
-      toMonth,
-      toDay,
-      rangeSeparator,
-    });
+    const { checked, fromValue, toValue } = this.state;
+    const { rangeSeparator, onApply, activeFilter } = this.props;
 
-    if (!dateRangeString) return;
+    if (!checked) return;
+    if (!fromValue && !toValue) return;
```
OliverGeneser: Does it correctly handle if the user inputs " "? (which is truthy)

src/lib/components/RangeFacet/RangeCustomFilter.js
```diff
@@ -70,213 +85,246 @@ class RangeCustomFilter extends React.Component {
...
-    const [fromISO, toISO] = dateRangeString.split(rangeSeparator);
-    onApply(
-      [DateTime.fromISO(fromISO).year, DateTime.fromISO(toISO).year],
-      dateRangeString
-    );
+    if (!this.validate()) return;
+
+    // Normalize to zero-padded ISO format (e.g. "2025-3" → "2025-03")
+    const normalizedFrom = this.normalizeValue(fromValue);
+    const normalizedTo = this.normalizeValue(toValue);
```
OliverGeneser, line 230: Should the values be normalized before we do the validation?

src/lib/components/RangeFacet/RangeCustomFilter.js
```diff
@@ -70,213 +85,246 @@ class RangeCustomFilter extends React.Component {
...
+              <Input
+                size="mini"
+                value={toValue}
+                onChange={this.handleChange("toValue")}
+                onFocus={this.activateOnFocus}
+                onKeyDown={this.handleKeyDown}
+                inputMode="numeric"
+                error={!!error}
+                className="searchkit-daterange-masked-input"
+              />
```
OliverGeneser, line 322: No love for the screen readers? (aria-label or can we use a htmlFor? or is that to new?)

src/lib/components/RangeFacet/RangeCustomFilter.js
slint: In theory you can't because the mask formatter filters out the space. If you meant empty string, then it also works. But I just noticed it doesn't work if you have both ranges being empty... which I think should at least show a validation error like "at least one end of the ranges is required" (I'll add this after the review).

src/lib/components/RangeFacet/RangeCustomFilter.js
slint: Fixed, there are props now for the radiobutton, both inputs, and the apply button.

src/lib/components/RangeFacet/RangeCustomFilter.js
```diff
@@ -70,213 +85,264 @@ class RangeCustomFilter extends React.Component {
...
-    const { fromYear, toYear, fromMonth, fromDay, toMonth, toDay } = this.state;
+  handleChange = (field) => (e) => {
+    const formatted = this.formatValue(e.target.value);
+    this.setState({ [field]: formatted, error: null });
+  };
 
-    if (String(fromYear).length !== 4 || String(toYear).length !== 4) {
-      return this.setError("Year must be 4 digits");
+  // Normalize a date value by zero-padding parts (e.g. "2025-3" → "2025-03")
+  normalizeValue = (value) => {
```
kpsherva, line 140: could the date normalisation, parsing and validation be done by an already existing library?

slint: `luxon` unfortunately doesn't do "partial" dates (e.g. `2025-3`). Then I wouldn't pull in an entire new dependency for just loosely parsing ISO dates.

After we normalize, I think we could use `luxon` for validation. The only caveat is that we lose the `Year must be 4 digits` custom validation we do.

So not sure we're gaining much, and I would prefer in the future we double-down and get rid entirely of Luxon and rely on the new [Temporal API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal).

kpsherva: can we then start introducing temporal from here?

slint: Probably not yet: https://caniuse.com/temporal (69% browser coverage)

Based on https://github.com/inveniosoftware/rfcs/blob/d162e8bf92fce8144439c4882af2f5d8fbbf2909/rfcs/rdm-0005-browser-support.md we could support only once it lands on stable supported versions of evergreen browsers...

